Issue #109, opened 2019-05-09
Gene association has been identified for a mitochondrial transporter


avlant:
### Description of the issue:
A mitochondrial amino acid transporter (ENSG00000164466) has recently been identified [1]. It transports serine and alanine, and should therefore be associated with HMR_5113 and HMR_5114.

[1] PubMed: 30442778,  https://www.uniprot.org/uniprot/Q9H9B4

#### Expected feature/value/output:
HMR_5113	m01307c  -> m01307m 		  0.00	1000.00	  0.00       ENSG00000164466
HMR_5114	m02896c  -> m02896m 		  0.00	1000.00	  0.00       ENSG00000164466

PR #114, opened 2019-05-22
feat: associate with new mitochondrial amino acid importer

Hao-Chalmers:
* New mitochondrial amino acid transporters were recently identified, as reported in #109. They are to  be included to `Human-GEM`, as well as the associated citation and proposed confidence score, using `associateMitoImporter_20190520.tsv`.
* At the mean time, the function `updateGrRules` is modified to allow the simultaneous update of `rxnReferences` and `rxnConfidenceScores` fields that are assumed to be provided whenever the gene-association is modified.


Hao-Chalmers:
@JonathanRob the reason for assigning confidence score was discussed under #109.

JonathanRob:
Ah, ok, I missed that part of the issue. That makes sense.

PR #118, opened 2019-05-27
Doc model changes

JonathanRob:

New functions `docModelChanges` and `writeModelChanges` are more generalized versions of the `docRxnChanges` and `writeRxnChanges` functions, respectively. These new functions now identify and document changes to metabolite-related fields in addition to reaction-related fields.

The `docModelChanges` and `writeModelChanges` functions help to automate documentation of changes to HumanGEM as it is curated. These functions will become less necessary as we migrate toward a binary-free repo (which would enable direct tracking of model changes), but are important in the mean time.


JonathanRob:
@Hao-Chalmers I agree that it's easier to assess the functions with some applications. I actually have these applications ready to go (will make a new PR soon); I just kept them separate to avoid a PR that made too many changes/additions.

Issue #107, opened 2019-04-22
feat: provide the stoichiometric matrix with comprehensive external association and annotation


Hao-Chalmers:
### Description of the issue:
* GEMs are mathematical representations of biochemistry and molecular biology. The GEM reconstruction and curation can be viewed as an iterative process of incorporating biological information from databases and literature.
* HumanGEM aims to provide comprehensive coverage of human metabolism, which is reflected through continuous curation of stoichiometric matrix.
* To increase scalability and robustness for the model, it is thus expected to comprehensively associate reactions and metabolites to external identifiers in major databases and GEMs.

#### Expected feature/value/output:
* Specify the `JSON` format for storing the association/annotation information (#75)
* Define a list of external database (e.g. `MetaNetX`, `KEGG`, `MetaCyc`) whose identifiers need to be provided
* Develop functions/scripts allowing regular curation of exteranl association/annotation, and consistency checks among inter-associated identifiers

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

Hao-Chalmers:
The Human1 reactions are annotated with external identifiers from **4** major databases (`KEGG`, `BiGG`, `REACTOME`, `MetaNetX`) and **2** previous Human GEMs (`EHMN`, `HepatoNET1`). This annotation/association has been provided in `Human1RxnAssoc.JSON` by commit [b13d86a](https://github.com/SysBioChalmers/Human-GEM/commit/b13d86a9549b9b911787bc515ec5fb6b2f8faf10).

Hao-Chalmers:
The Human1 metabolites are annotated with external identifiers from **7** major databases (`KEGG`, `BiGG`, `HMDB`, `ChEBI`, `PubChem`, `LipidMaps`, `MetaNetX`) and **3** previous Human GEMs (`EHMN`, `HepatoNET1`, `Recon3D`). This annotation/association has been provided in `HumanGEMMetAssoc.JSON` by commit [a2a60cc](https://github.com/SysBioChalmers/Human-GEM/pull/120/commits/a2a60cc91ee3176cd15cfb709d3bfb76bc67e212).

Hao-Chalmers:
Now the stoichiometric matrix annotation collected from various external sources has been centralized into **2** JSON files, which should be synced with the model structure (i.e. `rxns` and `mets` fields) whenever HumanGEM stoichiometric matrix is modified.

Issue #75, opened 2019-01-08
Apply JSON files for storing curation/annotation information


Hao-Chalmers:
### Description of the issue:
* This issue is derived from another ongoing issue (#27), which aims to drive the repo toward binary-free mode eventually
* [JSON](https://www.json.org) is an open-standard and language-independent file format. It is human-readable text and commonly used for data interchange
* JSON files appear to be ideal for archiving information in GEM-type repo and helping the migration to binary-free mode, given that Matlab recently started to support JSON by providing built-in functions (`jsondecode`, `jsonencode`) for import/export.

#### Expected feature/value/output:
* It is attempted here to utilize JSON files for storing the model annotation information and curation results that are often `cell array` type variables in **humanGEM** and previously saved as `.mat` files
* Develop necessary tailored functions (e.g. `prettyJson`) to support the smooth application of JSON files
* Utilization examples are given below, and will be updated if modified

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

Hao-Chalmers:
Examples for loading/exporting `cell array` structure from/to a JSON file:

 * Load JSON file `abc.json` to a variable:

```
variableName = jsondecode(fileread('abc.JSON'));
```

 * Export a variable to a `abc.json` file to :

```
jsonStr = jsonencode(variableName);
fid = fopen('abc.json', 'w');
fwrite(fid, prettyJson(jsonStr));
fclose(fid);
```

PR #120, opened 2019-05-31
feat: extensively provide associated external identifiers for rxns and mets

Hao-Chalmers:
* This PR is to address #107.
* Scripts `getHuman1RxnAssoc` and `getHuman1MetAssoc` are made for refining/retrieving previously prepared association/annotation to reactions and metabolites, respectively.
* **6** identifiers from different databases or human GEMs were assigned to reactions, and **10** identifiers for metabolites.
* The sorted external identifiers are saved to `HumanGEMRxnAssoc` and `HumanGEMMetAssoc`, both are organized in **JSON** format, as discussed in #75.


JonathanRob:
Ok, it will be very nice to have all the external ID information organized as you have done here. I have only two small comments:

1) I'm not sure if you're planning to merge the already generated JSON files directly, or regenerate the JSON files after merging. If you are going to merge them directly, just make sure that they're updated. The reason I'm saying this is because the current `humanGemRxnAssoc.JSON` file does not appear to be updated with your latest Recon3D ID additions.

2) How about `humanGEMRxnAssoc.JSON` instead of `humanGemRxnAssoc.JSON` (and same with the metabolite file)? Since GEM is the normal usage of the term, I would recommend using it that way.

Hao-Chalmers:
@JonathanRob can you specify any "latest Recon3D ID additions" so that they can be retrieved if missing.

JonathanRob:
@Hao-Chalmers What I meant was that when I ran the `getHuman1RxnAssoc` script, the new `humanGemRxnAssoc.JSON` file that it generated did not match the existing `humanGemRxnAssoc.JSON` file. It seemed that this was because the existing JSON file was an older version that you had generated before your commit (7a40123) `fix: include Recon3D ids into rxn association`. However, it looks like I ran it right before you updated the file in your next commit, so the issue has been solved.

PR #122, opened 2019-07-18
Annotation of model rxns, mets, and genes

JonathanRob:
The main component of this PR is a new function `annotateModel` that adds reaction, metabolite, and/or gene annotation information (i.e., alternate/external IDs) to the model. 

The function retrieves reaction and metabolite annotation information ifrom the `humanGEMRxnAssoc.JSON` and `humanGEMMetAssoc.JSON` files, respectively, whereas gene annotation information is retrieved with the help of the `translateGrRules` function, which uses the `ensembl_ID_mapping.tsv` file.

The retrieved annotation information is added to the corresponding model MIRIAM fields (`rxnMiriams`, `metMiriams`, `geneMiriams`) and/or as individual model fields (e.g., `rxnKEGGID`, `metChEBIID`, etc.).

### Other changes in this PR:
**Renaming MetaNetX fields**
Here it is proposed to re-name the `rxnMNXID` and `metMNXID` fields to `rxnMetaNetXID` and `metMetaNetXID`, respectively, to be consistent with how these fields are named in COBRA-style models. These MNX fields in the `humanGEMRxnAssoc.JSON` and `humanGEMMetAssoc.JSON` annotation files were therefore renamed as such.

### DIFF
--- a/ComplementaryData/annotation/humanGEMMetAssoc.JSON
+++ b/ComplementaryData/annotation/humanGEMMetAssoc.JSON
@@ -111143,7 +111143,7 @@
 		"",
 		"",
 		""],
-	"metMNXID":[
+	"metMetaNetXID":[
 		"MNXM45735",
 		"MNXM45735",
 		"MNXM163755",
--- a/ComplementaryData/annotation/humanGEMRxnAssoc.JSON
+++ b/ComplementaryData/annotation/humanGEMRxnAssoc.JSON
@@ -94646,7 +94646,7 @@
 		"",
 		"",
 		""],
-	"rxnMNXID":[
+	"rxnMetaNetXID":[
 		"MNXR95725",
 		"MNXR95726",
 		"MNXR95413",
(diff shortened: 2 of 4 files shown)


Hao-Chalmers:
The purpose of this `annotateModel` is not very clear to me. So on what conditions it will be used?

JonathanRob:
Very good question! 

We have a lot of annotation information for reactions, metabolites, and genes that is somewhat distributed among different files (e.g., `humanGEMMetAssoc.JSON`), and not directly integrated in the HumanGEM structure.

This function is an alternative to permanently adding all of the annotation information to HumanGEM as additional structure fields. I am avoiding this option to permanently add these fields, because:
1. It increases the size (filesize and complexity) of the model with content that may not be relevant for many users.
2. Many functions (e.g., in RAVEN or COBRA toolboxes) do not properly update these fields when editing the model. For example, if I delete a reaction, it may not delete its corresponding entry in the `rxnMetaNetXIDs` field.

In some cases, users may want to have this information contained directly in the fields of the model structure, as it is in many other GEMs. For example, if I want to export the model to SBML for use in other applications, I may want to have the additional fields as part of the SBML file. The `annotateModel` function thus enables users to very easily add all this information to the model in a **standardized** format (i.e., proper formulation in the MIRIAM fields).

Hao-Chalmers:
This `annotateModel` function should be useful in some conditions, though it should be noted that the MIRIAM fields (`rxnMiriams`, `metMiriams`, `geneMiriams`) might be excluded from RAVEN/COBRA model structure soon.

JonathanRob:
@Hao-Chalmers Yes, it is true that the MIRIAM fields will be removed eventually, so I plan to update the function in the future to not add these fields by default.

Hao-Chalmers:
ok, merged.

Issue #117, opened 2019-05-27
Extracellular metabolites without exchange reactions


avlant:
### Description of the issue:
There seems to be 4 extracellular metabolites that lack exchange reactions.

 20-hydroxy-arachidonate[Extracellular] 
 chenodiol[Extracellular] 
 LacCer pool[Extracellular] 
 Chylomicron Lipoprotein[Extracellular] 

Since they are outside of the cell they should be able to escape over the boundary.

#### Expected feature/value/output:
I would expect that each extracellular metabolite has an exchange reaction.

Hao-Chalmers:
Here are the corresponding met ids:

`m00591`:   20-hydroxy-arachidonate
`m01435`:   chenodiol
`m02328`:   LacCer pool
`chylo_hs`:   Chylomicron Lipoprotein

avlant:
@Hao-Chalmers I think this is all. I compared all metabolites in the extracellular compartment with the metabolites associated with exchange reactions identified using getExchangeRxns(model); But the test can be written as a single line:
setdiff(model.metNames(model.metComps == 1), model.metNames(model.metComps == 9))

JonathanRob:
I also ran another check and found the same set of metabolites. I've written a script to implement these exchange reactions into the model, but I will first wait until PR #113 is incorporated before pushing the files.

Hao-Chalmers:
@JonathanRob okay, let's hold this until #113 is done.

JonathanRob:
Now PR #123 has added new exchange reactions for the 4 metabolites listed above. In addition, five reactions (`HMR_9736`, `xenobiotics`, `arachidonates`, `steroids`, and `others`) were found to involve transport between the boundary and a non-extracellular compartment. 

These reactions were revised to transport to the extracellular compartment, and a new reaction was added for those that did not have a corresponding exchange reaction from extracellular to boundary.

Hao-Chalmers:
A

PR #123, opened 2019-08-01
Add exchange rxns

JonathanRob:

[Addresses Issue #117]

Exchange reactions are added for four metabolites that are present in the extracellular compartment, but do not currently have exchange reactions transporting them to/from the boundary. These changes are implemented with the `curateExchangeRxns2.m` script.

The four metabolites for which exchange reactions are added:
1)  20-hydroxy-arachidonate
2)  chenodiol
3)  LacCer pool
4)  Chylomicron Lipoprotein

In addition, the script addresses the issue of five reactions that involve transport between the boundary compartment and a non-extracellular compartment. The reaction IDs are:
1)  HMR_9736
2)  xenobiotics
3)  arachidonates
4)  steroids
5)  others

Information on the new reactions added to the model is contained in the `newExchangeRxns.tsv` file.


JonathanRob:
This PR also demonstrates the use of the new `docModelChanges` and `writeModelChanges` functions introduced in PR #118.

Hao-Chalmers:
@JonathanRob this PR looks good.

Would it be better to name the files according to the targeted issue? For example, change filename from `newExchangeRxns.tsv` to `newExchangeRxns_issue117.tsv`, and `curateExchangeRxns2.m` to `curateExchangeRxns_issue117`.

JonathanRob:
@Hao-Chalmers I think that's a good idea. It would make it quite clear which files are associated with which issues/changes.

Hao-Chalmers:
This PR actually adds `8` exchange rxns that are claimed for addressing #117, where there are only 4 exchange rxns were proposed. It would thus be necessary to also mention the other changes (mets and exchange/transport rxns) in #117.

JonathanRob:
@Hao-Chalmers Yes, this PR adds more exchange reactions than the 4 discussed in #117, but these are added for a different reason (as explained in the description of this PR).

The 4 exchange reactions discussed in #117 are to allow exchange of metabolites that are in the extracellular compartment, but do not currently have exchange reactions to/from the boundary. The other exchange reactions added in this PR are for a separate problem, where metabolites are being transported to/from the boundary from a non-extracellular compartment. Since this was not the topic of #117, I don't think there's a need to discuss these other reaction changes there.

Issue #124, opened 2019-08-09
Invalid BiGG IDs


JonathanRob:
### Description of the issue:
Many metabolite and reaction BiGG IDs in the association files (`humanGEMMetAssoc.JSON` and `humanGEMRxnAssoc.JSON`) are invalid, meaning they do not correspond to an ID in the BiGG database (version 1.5 at the time this issue was created).

For example, `ala_L` is the BiGG ID associated with L-alanine. However, the correct BiGG ID is `ala__L`.

This seems to have largely been due to the use of invalid BiGG IDs in Recon3D, from which many of HumanGEM BiGG IDs originated. Others are old or expired BiGG IDs, errors carried over from previous reconstructions, and other miscellaneous sources.

Regardless of the cause, all BiGG IDs in the model annotation files should either be valid, or removed.

#### Expected feature/value/output:
Association files with 100% valid metabolite and reaction BiGG IDs.

#### Current feature/value/output:
Based on comparison with BiGG database v1.5:
- Invalid/old metBiGGIDs: 2,874
- Invalid/old rxnBiGGIDs: 687

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Checked that a similar issue does not exist already

Hao-Chalmers:
@JonathanRob how many invalid BiGG ids were found here, if you have the specific numbers?

JonathanRob:
- Invalid/old metBiGGIDs: 2,874
- Invalid/old rxnBiGGIDs: 687

Hao-Chalmers:
Since the BiGG database has been continously changing and regularly updated, it would be better to explicitly indicate the database version onto the issue description, together with these invalid numbers.

JonathanRob:
Good point. I have updated the description to include the database version (1.5) as well as the numbers I provided above.

JonathanRob:
These BiGG ID updates have now been implemented (PR #135), so this issue can be considered solved.

PR #125, opened 2019-08-09
Fix BiGG IDs

JonathanRob:
This fix addresses Issue #124, which states that the model annotation files `humanGEMMetAssoc.JSON` and `humanGEMRxnAssoc.JSON` contain many BiGG IDs that do not correspond to any IDs on the BiGG Database.

The `updateBiGGIDs` script corrects/updates the metabolite and reaction BiGG IDs in the annotation files. The only IDs that could not be successfully updated are metabolite `eumelanin`, and reactions `DM_Asn-X-Ser/Thr(ly)` and `sink_pre_prot(er)`, for which BiGG IDs do not exist (that properly correspond to the metabolite/reaction with which they are associated).


Hao-Chalmers:
Since this PR doesn't modify the model and only modify the `JSON` annotation files, which can be directly updated here.

PR #130, opened 2019-09-18
Curate lipid pools

JonathanRob:
This PR implements changes in the way that HumanGEM handles many of its lipid pools, such as `fatty acid-LD-PC pool`, `1-acylglycerol-3P-LD-SM pool`, `acyl-CoA-LD-PS pool`, etc.

These pools are currently formulated in such a way that their composition differs, as described in the original HMR2 [paper](https://doi.org/10.1038/ncomms4083) (specifically in Supplementary Data 4). The original motivation was to have the pool compositions reflect those from plasma and liver tissue measurements. In practice, this leads to stoichiometric inconsistencies, because the model incorrectly treats these pool metabolites have having equal mass.

To address this problem, the pool metabolites involved in the lipid pool formation and degradation reactions were simply renamed to a specific "liver" lipid pool, which will essentially disconnect them from other reactions in the model. New lipid pool formation/degradation reactions are then added, whose compositions represent an **average** fatty acid composition of these different pools.

These pool compositions now enable a more generic treatment of the pool metabolites, resulting in a more stoichiometrically consistent model that still utilizes experimental fatty acid composition data.

The changes are implemented in the `curateLipidPools` script, which updates HumanGEM and the metabolite and reaction annotation files accordingly.

**Other changes in this PR**
The following three now-deprecated fields are removed from the `HumanGEM` model structure in the same `curateLipidPools` script:
1. proteins
2. prRules
3. rxnProtMat

The protein-related fields can simply be regenerated using the `translateGrRules` function.


JonathanRob:
This PR should not be merged until **after** #123.

JonathanRob:
Ok, after some small revisions and re-checking, this PR is now ready.

Hao-Chalmers:
I completely agree to remove these four fields. But for the moment it might be better to keep this `priorCombiningGrRules`, which could be useful in some validation analysis.

JonathanRob:
@Hao-Chalmers Ok, I have revised to maintain the `priorCombiningRules` field, but this field should be removed very soon.

Issue #27, opened 2018-09-24
Migrating toward a binary-free repo


Hao-Chalmers:
### Description of the issue:
* GitHub is a versioning system for software and databases that stores mainly text files, because their modifications can be easily traced and controlled. Changes made to binary files are invisible so  they are generally avoided in git repositories.
* In GEM-type repos, `.mat` files are the common format used in model reconstruction and simulation analysis and thus an inevitable component. However, `.mat` files are binary and immediately lead to conflicts if modified in different branches and merged together. Therefore, abandonment of `.mat` files had been proposed and implemented, such as in [yeastGEM repo](https://github.com/SysBioChalmers/yeast-GEM/issues/49).
* One major object of hosting GEMs as repositories is to provide public framework contributed by and serving for the research community through an open and transparent curation procedure. The `.mat` files contradict this purpose since commits applied to them are unable to track.

#### Expected feature/value/output:
* This issue aims to develop an workable plan for migrating the repo into a binary-free mode.
* It is certain that the migration will take some time.
* At the moment, all changes to `.mat` file is restricted only to `devel` branch with coordinations between contributors.

**I hereby confirm that I have:**
- [x] Checked that a similar issue does not exist already

Hao-Chalmers:
This migration has been achieved with #156.

PR #132, opened 2019-09-20
feat: enable lossless conversion between `mat` and `yml` model formats

Hao-Chalmers:
* This PR aims to allowing for a binary-free workflow in model curation, as addressed in #27
	* Add new `importHumanYaml` function to enable loading models directly from `Yaml` format, which was proposed by [cobrapy](https://cobrapy.readthedocs.io/en/latest/io.html#YAML) and is light-weight compared to `SBML` format
	* Modify `writeHumanYaml` function to ensure lossless conversion between Matlab and Yaml model formats
	* Add `testYamlConversion` function for testing if the conversion is successful or not


simas232:
I ran `testYamlConversion` in my machine and got a negative result (`status=0`). Here are most visible differences between the compared models:

| Field Name | `HumanGEM` | `importedHumanGEM` |
|:-------------:|:------------:|:---------------:|
|  `rev`<br>`c` |     13520x1 double     |  13520x1 logical   |
|	 `compOutside`     |      10x1 cell   |  0x0 cell   |
|	 `unconstrained`     |      10103x1 double   |  10103x1 logical   |
|	 `metCharges`     |      10103x1 int64   |  10103x1 double   |

Upon removing these fields, the models were still not identical. There were differences in `rxnReferences` for the instances where double quotation marks were used, for example:
`HumanGEM.rxnReferences{9742,1}` has double quotation marks while `importedHumanGEM.rxnReferences{9742,1}` does not.

Hao-Chalmers:
@simas232 very good reviews/comments, which led to the corrections of some field types.
This `testYamlConversion` function only takes care the shared fields, because the non-shared ones will be abandoned (e.g. `compOutside`) in the future.
Among the shared ones, `rev` and `rxnReferences ` are not identical. The former is due to errors and the latter is caused by quotation marks, these issue will have to be fixed when implementing the yml-based curation workflow.

Hao-Chalmers:
> The functions importYaml and getGenesFromGrRules are very welcome in RAVEN.

Yes, these functions will be migrated to RAVEN after some testing in this repo and figuring out a feasible plan about the `metadata` section.

Hao-Chalmers:
Now the annotation sections in the HumanGEM `yml` file are not fully compatible with the specs defined in Cobrapy. This will be fixed in the future developments.

ComplementaryScripts/Functions/importHumanYaml.m
```diff
@@ -0,0 +1,307 @@
...
+%   Hao Wang, 2019-07-26
+%
+% This function is to reverse engineer the RAVEN function `writeYaml`
+%
+
+if ~(exist(yamlFilename,'file')==2)
+    error('Yaml file %s cannot be found',string(yamlFilename));
+end
+
+% Define the required fields of humanGEM
```
mihai-sysbio: I'd consider doing this assignment on a single line.

ComplementaryScripts/Functions/importHumanYaml.m
```diff
@@ -0,0 +1,307 @@
...
+    % import different sections
+    
+    % import metaData
+    if isequal(tline, '- metaData:')
+        fprintf('\tmetaData\n');
+        section = 1;
+    end
+
+    if section == 1 && numel(tline) > 17
+        if isequal(tline(1:17),'    short_name  :')
```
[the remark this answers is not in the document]
Hao-Chalmers: Sounds a good idea since the metaData section is actually rigidly defined. Will be implemented in next round of refinement.

ComplementaryScripts/Functions/importHumanYaml.m
Hao-Chalmers: sorry I don't understand.

mihai-sysbio: Something like this:
`[model.id, model.description] = deal([], [])`
`[model.rxn, model.mets] = deal({}, {}) `

PR #133, opened 2019-09-23
style: capitalize model names

Hao-Chalmers:
* This PR is to capitalize all model files in the repo
	* Two model filenames in the `devel` branch are capitalized
	* All functions for model releasing, exporting and routine manipulating are adjusted for using or generating capitalized model filenames.


mihai-sysbio:
It might be easier if the repository name would be the exact same name as the repo (including `-` between `Human` and `GEM`).

PR #135, opened 2019-10-04
Human 1.1.2

Hao-Chalmers:
- Features (PR #132):
	- New function `importHumanYaml` and modified `writeHumanYaml` for enabling lossless conversion between `mat` and `yml` model files
- Fixes (PR #125):
	- Systematically correct invalid BiGG rxn/met ids in `humanGEMMetAssoc.JSON` and `humanGEMRxnAssoc.JSON`, as detailed in #124


JonathanRob:
Maybe it's too late for this suggestion, but for the functions `importHumanYaml` and `writeHumanYaml`, I would recommend renaming one of them so they are `importHumanYaml` and `exportHumanYaml`, or `readHumanYaml` and `writeHumanYaml`. This would be more intuitive and perhaps easier to remember these functions.

Hao-Chalmers:
@JonathanRob sure they can be renamed when implementing.

Note that the `export` prefix had been taken by another `exportHumanGEM` function, and the current way import/write was actually adapted from RAVEN.

PR #140, opened 2019-10-15
Further iHsa integration

JonathanRob:

This PR performs a more extensive integration of **iHsa** (EM Blais et al. _Nat Commun_ 2017) content into **HumanGEM**. Specifically, the following is added to **HumanGEM** from **iHsa**:
- 67 new reactions
- 60 new metabolites
- 688 updated KEGG reaction associations
- 1 new gene

Although there were ~100 additional reactions in iHsa that were not present in HMR2 (and therefore also not in Human-GEM), these were excluded because they were either mass imbalanced, part of a mass-imbalanced pathway, or involved the transport or export of a metabolite that was not present in Human-GEM.

In addition, a new `rxnRatconID` field was added to the `humanGEMRxnAssoc.JSON` file to provide association information between HumanGEM and iHsa.


JonathanRob:
Note that in the `iHsaAdditionalIntegration` script I have now deprecated (removed) the `priorCombiningGrRules` field.

PR #144, opened 2019-10-28
fix: remove inchi entries that conflict with metFormulas

JonathanRob:
Many of the entries in the `inchis` field are out of sync with the `metFormulas` field (see issue #145).

This PR introduces a new function `removeConflictingInchiStrings` that compares all non-empty `inchis` entries with their corresponding entry in `metFormulas` and deletes the `inchis` entry if they disagree. This function (or some variant) could be re-run on the model anytime the `metFormulas` or `inchis` fields are modified to ensure that they remain synced.

The model can be updated by simply running the following command:
`ihuman = removeConflictingInchiStrings(ihuman);`

### DIFF
--- a/ComplementaryScripts/modelCuration/removeConflictingInchiStrings.m
+++ b/ComplementaryScripts/modelCuration/removeConflictingInchiStrings.m
@@ -1,22 +1,24 @@
-function updated_model = updateModelInchiStrings(model)
+function corrected_model = removeConflictingInchiStrings(model)
 % Removes inchi strings that conflict with their corresponding metFormula
 %
 % Input:
 %
-%   model           model structure
+%   model   model structure
 %
 %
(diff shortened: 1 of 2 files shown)


Hao-Chalmers:
The function is to fix the mismatch between formulas and `inchi` strings in a model, which shouldn't be a common situation. Maybe just turn it into a fixation script?

JonathanRob:
In theory, no, it shouldn't be a common situation. However, it could easily happen anytime someone changes `metFormulas` and forgets to update the corresponding `inchi` field. Maybe just including a check somewhere that these two fields are consistent?

Hao-Chalmers:
Of course it's okay to have such a function for this purpose, then would be better to name it more specificly, e.g. `removeConflictInChIstr`.

PR #156, opened 2020-02-04
feat: activate yaml workflow

Hao-Chalmers:
* This PR is to start the implementation of yaml-based curation workflow, as planned in #27
	* After confirmation of lossless conversion between Yaml and Matlab model files (using `testConversionYaml` function), the latter is removed from current and later all other branches, except `master`
	* From now on, a yaml-based workflow is thus activated and simple curations can be made by directly editing model components (e.g. subSystems, EC numbers) with the Yaml file  
	* Complex curations can be made by functions/scripts after loading the model (using `importHumanYaml`) from, and finally save back (using `writeHumanYaml`) to, the Yaml file
	* Minor fixes to `importHumanYaml` and `testConversionYaml`
	* It is suggested that the Yaml file should be checked by `testConversionYaml` every time when merging feature/fix branches into `devel`


JonathanRob:
For some reason, changing the `.yml` file (i.e., in e4b1093) seems to generate a massive diff (57,588 lines changed). I would suggest that this is fixed before implementing this type of workflow, because:
1. Huge diffs will bloat the repo (I think?)
2. It's virtually impossible to see what was actually changed in the yaml file, which completely defeats its entire purpose.

Hao-Chalmers:
@JonathanRob thanks for spotting this huge diffs problem, which was caused by Atom editor that introduced some spaces. So it need to be careful when editing Yaml in text editors.

Here I pushed another commit that fixed the problem.

PR #159, opened 2020-03-12
Reorganize repository directory structure

JonathanRob:
The directory structure is reorganized and the name of some directories are changed in an effort to make functions more intuitive to find, and to be more consistent with [standard-GEM](https://github.com/SysBioChalmers/standard-GEM). This would ideally be implemented before publication of the manuscript, as it is a potentially disruptive change.

- The `complementaryData/` directory was renamed to `data`.
- The `complementaryScripts/` directory was renamed to `code`.
- The `code` directory now has the following structure:
```
code
├── GPRs
├── io
├── misc
├── modelCuration
│   ├── GPRs
│   │   └── EnzymeComplexes
│   ├── MetAssociation
│   ├── RxnAssociation
│   └── modelIntegration
├── qc
└── tINIT
```

### Proposed deprecation of curation scripts
Many of the curation scripts were necessary to help keep track of changes made to the model `.mat` file because it is binary, and/or were implemented for a one-time curation action. As such, their only remaining purpose is for transparency and re-tracing the steps of the curation process. All such scripts have been moved to the `code/modelCuration/` subdirectory.

Since these scripts will forever remain in past versions of this repository but are no longer in use, I propose that we plan to **eventually remove these scripts from future repository instances**. Therefore, if any scripts/functions in the `code/modelCuration/` directory are still useful, they should be moved to another appropriate location in the `code/` directory to avoid removal.

#### Deletion of two previously deprecated functions
For some reason, the `docRxnChanges` and `writeRxnChanges` functions are still present, even though they were replaced by the more general `docModelChanges` and `writeModelChanges` functions, respectively. These functions have now been removed.


JonathanRob:
I am open to comments/suggestions on the new directory structure I've proposed here. Also, GitHub is unfortunately marking several scripts as deleted and added, even though the only change was to their path. Suggestions to help with a cleaner diff are appreciated.

mihai-sysbio:
180 file changes are _pure_ renames, 2 files are fully deleted and the rest seem to be both renamed and modified (path changes count like renames). My only suggestion is to save the overview above into `code/README.md`.

Hao-Chalmers:
This change is according to the `standard-GEM`, which should be mentioned somewhere.

JonathanRob:
Ok, I've added a reference to [standard-GEM](https://github.com/SysBioChalmers/standard-GEM), and fixed path references so that no functions/scripts are disrupted by the new directory structure.

I've also added a few small changes to the `HumanGEM.yml` annotation:
- email has been updated to my email address
- the model description was slightly revised
- non-English (MATLAB-incompatible) characters were replaced in Pınar's name

pinarkocabas:
Thank you  very much

Pınar Kocabaş
Chalmers University of Technology
Department of Biology and Biological Engineering
Division of Systems and Synthetic Biology
Kemivägen 10,
SE-412 96
Göteborg, Sweden

________________________________

PR #160, opened 2020-03-26
fix: adjust increaseHumanGemVersion and other functions

Hao-Chalmers:
* Modify `increaseHumanGemVersion` function to adapt with the Yaml-based workflow and enable `xlsx` format output
* Fix path in `annotateModel` function
* Adjust `README` according to changes made to path and files


code/annotateModel.m
```diff
@@ -201,7 +201,7 @@
         for i = 1:numel(model.mets)
             if hasMet(i)
                 % add annotation
-                model.metMiriams{i} = appendMiriamData(model.metMiriams{i}, miriamNames, miriamValues(metInd(i),:)');
+                %model.metMiriams{i} = appendMiriamData(model.metMiriams{i}, miriamNames, miriamValues(metInd(i),:)');
```
JonathanRob: Is there a reason the `metMiriams` field is now omitted?

PR #167, opened 2020-04-20
fix: remove dash from model ID while exporting to SMBL format

JonathanRob:
The dash in the model ID `Human-GEM` causes problems with SBML I/O, so this fix removes the dash (`HumanGEM`) in **only the SBML file**. The model ID will remain as `Human-GEM` everywhere else.


mihai-sysbio:
Nice fix! Would it make sense to also rename the `.yml` file to Human`-`GEM.yml?

JonathanRob:
@mihai-sysbio Yes, it would probably be a good idea to change the model filenames as well to be consistent. We just need to make sure that the names of all other model versions (`.xml`, `.mat`, etc.) and all functions reading/writing these files are also updated.

Hao-Chalmers:
yes, we could add a project "Aligning model filenames with ID of repo and model" to coordinate efforts toward this goal, probably through multiple commits/PRs.

JonathanRob:
@Hao-Chalmers I think creating an issue is sufficient, it's not a big enough change to be considered a "project."

Issue #153, opened 2019-12-18
Remove HMR_4362 from the model


avlant:
### Description of the issue:
The following reaction HMR_4362 (ATP[c] + H2O[c] + pyruvate[c] => AMP[c] + PEP[c] + Pi[c]) with EC number 2.7.9.2 is not supported by a reference in the model, and does not seem to exist in human according to KEGG [1], Brenda [2] or in textbook descriptions of gluconeogenesis [3, 4]. 

[1] https://www.genome.jp/kegg-bin/show_pathway?org_name=hsa&mapno=00620&mapscale=&show_description=hide
[2] https://www.brenda-enzymes.org/enzyme.php?ecno=2.7.9.2
[3] https://socratic.org/questions/how-is-gluconeogenesis-related-to-glycolysis
[4] https://themedicalbiochemistrypage.org/gluconeogenesis.php

#### Expected feature/value/output:
The reaction should be removed.

#### Current feature/value/output:
It exists.

Hao-Chalmers:
@avlant thanks for posting the issue.

If you happen to know the reason why this reaction was included initially, maybe that would help in avoiding similar problems in the future.

JonathanRob:
Addressed in #171.

mihai-sysbio:
@all-contributors please include @avlant for bug and content

mihai-sysbio:
Alright, let me try again.

@all-contributors please add @avlant for bug and content

Issue #129, opened 2019-09-11
Remove Nitrogen metabolism subsystem


JonathanRob:
### Description of the issue:
The model currently contains a subsystem `Nitrogen metabolism` that has only one reaction:

HMR_5407: `CO2 +  H2O <=> carbonate`

This reaction should be merged into another subsystem, and the `Nitrogen metabolism` subsystem should be removed.

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

JonathanRob:
In addition, the `Metabolism of Other Amino Acids` and `Metabolism of other amino acids` subsystems (differing only in capitalization) should be merged.

JonathanRob:
Addressed in #171.

PR #171, opened 2020-05-21
Fix: subsystem changes and reaction deletion

JonathanRob:
This PR makes a few small changes to address Issues #129 and #153.

#### Subsystem changes
Addresses Issue #129 

Reaction `HMR_5407: CO2[s] + H2O[s] <=> carbonate[s]` was the only reaction assigned to the `Nitrogen metabolism` subsystem. The subsystem for this reaction was changed to `Miscellaneous`, the same as the other compartment versions of this reaction (`HMR_5406` and `HMR_8751` in cytosol and mitochondia, respectively). This effectively removes the one-reaction `Nitrogen metabolism` subsystem.

In addition, the capitalization of the `Metabolism of Other Amino Acids` subsystem was changed to `Metabolism of other amino acids`, to merge these two subsystems.

#### Reaction deletion
Addresses Issue #153 

Based on the provided evidence (see Issue #153) and further searching, the following reaction:
`HMR_4362: ATP +  H2O +  pyruvate =>  AMP + 2 H+ +  PEP +  Pi`
is not known to occur in humans (it is microbial), and was therefore removed from the model.


JonathanRob:
**Note**: I suggest we first merge #173 so I can update this branch with the new version of `devel`, then merge this branch.

Hao-Chalmers:
@JonathanRob I agree to first merge #173, which is more heavyweight.

JonathanRob:
The changes have been re-implemented after merging #173, and should now be ready for merge with `devel`.

Issue #169, opened 2020-05-10
YAML format issues in modelFiles/yml/HumanGEM.yml


dweindl:
Hi, great project! I was trying to look into the yaml model, but ran into some issues:  

### Description of the issue:

YAML errors in `modelFiles/yml/HumanGEM.yml` preventing parsing the file.

#### Expected feature/value/output:

Expected to be able to load `modelFiles/yml/HumanGEM.yml` with any yaml parser.

#### Current feature/value/output:

Trying to load `modelFiles/yml/HumanGEM.yml` fails.

Main issue is using ":" in unquoted strings, leading to 

> syntax error: mapping values are not allowed here (syntax)

as well as long list of other issues which seem to be minor.

#### Reproducing these results:

```shell
yamllint modelFiles/yml/HumanGEM.yml
```
-> long list of issues

or 

```python
import yaml
yaml_file = "modelFiles/yml/HumanGEM2.yml"
with open(yaml_file) as f:
    model = yaml.safe_load(f)
```
-> `[...] yaml.scanner.ScannerError: mapping values are not allowed here`

**I hereby confirm that I have:**
- [x] Tested my code on my own computer for running the model
- [x] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

Hao-Chalmers:
@dweindl thank you posting the issue.

The parsing error is because the specification of "HumanGEM.yml" is slightly different from the one defined in Cobrapy. The HumanGEM Yaml format is originated from RAVEN, and later adjusted according to specific needs (#27, #71).

One can use either the SBML "HumanGEM.xml" or MATLAB "HumanGEM.mat" file for importing model.

dweindl:
Thanks for the quick reply, @Hao-Chalmers . The issue occurs independently of cobrapy. The current file is not valid yaml. I think most of it would be fixed by quoting strings during yaml export in 

https://github.com/SysBioChalmers/Human-GEM/blob/74de28c4b03a4e147a2e62a9ba894b981d6ddd75/code/io/writeHumanYaml.m#L234 

the same way as in 

https://github.com/SysBioChalmers/Human-GEM/blob/74de28c4b03a4e147a2e62a9ba894b981d6ddd75/code/io/writeHumanYaml.m#L232

Hao-Chalmers:
Can you please confirm if this error can be fixed by systematically adding double quotes?

dweindl:
[...]

Issue #77, opened 2019-01-21
Naming convention for rxns/mets identifiers in humanGEM


Hao-Chalmers:
### Description of the issue:
* There have been no newly added reactions in **humanGEM** since the generation. Yet new reactions (and mets) will come now and then later. And the naming convention of forthcoming rxns/mets ids should be determined in advance.
* Here I propose a nomenclatures that is expected to be applied to new reaction/metabolite identifiers in the future.
* This issue is to bring up open discussion to the proposal. Constructive suggestions and ideas are highly welcome.

#### Expected feature/value/output:
* All existing **HMR** rxns/mets ids will remain unchanged, to avoid any further conversions.
* Given that the **humanGEM** is inherited from **HMR**, it is preferred to follow the existing style for new ids, simply by increasing the number of numerical part.
* For example, **HMR** has rxn ids from `HMR_0001` up to `HMR_9818`. I suggest to assign new reactions with ids starting form `HMR_10000`, and `HMR_10001`, `HMR_10002`, and onward...
* The existing met ids were organized as such: 'm' is followed by 5 digits and ended with comp abbreviation (e.g. `m00001c`). So far, the middle five digits range from 00001 up to 03170. Therefore, the upcoming new mets could be assigned as `m03171c`, `m03172m`, and onward...
* This would also reserve the identify of **HMR** database, despite of its renaming to **humanGEM**

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository v0.7.0
- [X] Checked that a similar issue does not exist already

vandutt:
Hello,
For the mets id in humanGEM, there are currently two types of assignation of compartments. Could m03171c be m03171_c, like tmacmp_c, to facilitate the parsing?
Thank you.
Du Tran

Hao-Chalmers:
@vandutt thank you for the comments.

HumanGEM is inherited from the `HMR2` and `Recon3D`, which have different met id formats that have been retained for backward compatibility.

Note that one can directly get the metabolite compartment info through the `metComps` field provided in the model.

Hao-Chalmers:
This convention for naming new rxn/met ids has been adopted since v0.8 and functioning well.

Issue #71, opened 2018-12-03
Metadata in plaintext exports


mihai-sysbio:
The plaintext model files (e.g. `yaml`) should include metadata. This should be at least the version number, but it could also be a list of authors, a url, and a short description.

Hao-Chalmers:
This issue had been well resolved in #95, and thus closed.

PR #173, opened 2020-05-27
feat: yaml worflow

mihai-sysbio:
This PR addresses all yaml indentation issues from #169 and adds a GitHub Action to run `yamllint` on PRs to `master` and `devel`, and pushes to `devel`. The `yamllint` configuration `{extends: default, rules: {line-length: disable}}`, which, unlike `extends: relaxed` means most problems will result in errors, [as per the documentation](https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration). The `line-length` is disabled because it's unlikely this will be addressed.
I've ran `testYamlConversion.m` on the new yaml but nothing more.

Edit: Cobrapy import has also been added to the workflow.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -2,9 +2,9 @@ name: YAML validation
 
 on:
   push:
-    branches: [ develop, feat/yaml-workflow ]
+    branches: [ devel ]
   pull_request:
-    branches: [ master, develop ]
+    branches: [ master, devel ]
 
 jobs:
   yaml-validation:
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -27,3 +27,12 @@ jobs:
         # format: # optional, default is colored
         ## Return non-zero exit code on warnings as well as errors
         # strict: # optional, default is false
+
+    - name: Install cobrapy
+      run: |
+        sudo pip install pip --upgrade
+        sudo pip install cobra
+
+    - name: Test cobrapy import
+      run: 
+        python -c "import cobra ; cobra.io.load_yaml_model('modelFiles/yml/HumanGEM.yml')"
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -18,13 +18,13 @@ jobs:
       uses: ibiqlik/action-yamllint@v1.0.0
       with:
         # File(s) or Directory, separate by space if multiple files or folder are specified
-        file_or_dir: modelFiles/yml/HumanGEM.yml
+        file_or_dir: modelFiles/
         
         # Path to custom configuration
         # config_file: # optional
         
         # Custom configuration (as YAML source)
-        config_data: # optional
+        config_data: "{extends: relaxed}"
         
         # Format for parsing output [parsable,standard,colored,auto]
         # format: # optional, default is colored
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -1,0 +1,33 @@
+name: YAML validation
+
+on:
+  push:
+    branches: [ develop, feat/yaml-workflow ]
+  pull_request:
+    branches: [ master, develop ]
+
+jobs:
+  yaml-lint:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    
+    - name: YAML Lint
+      uses: ibiqlik/action-yamllint@v1.0.0
+      with:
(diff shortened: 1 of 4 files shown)


Hao-Chalmers:
@mihai-sysbio it's very nice to have GH actions checks!

Hao-Chalmers:
There is still one warning reported from yamllint and maybe worth to fix it here?
`1:1       warning  missing document start "---"  (document-start)`

mihai-sysbio:
I saw that, but was afraid to fix it - I'm not sure if _all_ yaml parsers can handle `---`.

JonathanRob:
It looks like the changes to the yaml indentation (d075b07) broke compatibility with cobrapy, unfortunately.

mihai-sysbio:
Compatibility with `cobrapy` has been restored, and is now also part of the same GH action.

mihai-sysbio:
Thanks @Hao-Chalmers and @JonathanRob for the feedback in resolving the PR.

BenjaSanchez:
@Hao-Chalmers @mihai-sysbio I found this PR and I wonder why are all these modifications to `writeHumanYaml.m` not a part of RAVEN's `writeYaml.m`? Moreover, any reason to have `writeHumanYaml.m` at all and not directly rely on `writeYaml.m`? That way all these changes could benefit all repos that uses `writeYaml.m`, for instance it would resolve failing tests at https://github.com/SysBioChalmers/yeast-GEM/pull/236

mihai-sysbio:
As you're suggesting @BenjaSanchez I also think it's better to fix problems at the source, in this case to apply the fixes directly to RAVEN. However, I'm not sure why [`writeHumanYaml`](https://github.com/SysBioChalmers/Human-GEM/commits/3783401ec8c9b662caf8fad1d22b2c9fa1435b5a/ComplementaryScripts/Functions/writeHumanYaml.m) came to be - maybe @JonathanRob remembers?

Hao-Chalmers:
@BenjaSanchez the development of `writeHumanYaml` was triggered by this issue #71.

JonathanRob:
The initial form of `writeHumanYaml` was super specific to Human-GEM and its parsing by Metabolic Atlas, so at that time it didn't make sense to apply the changes directly to RAVEN. Now that it has developed more, maybe it has reached a point that it is feasible to work toward merging it with the RAVEN function.

mihai-sysbio:
That's my mistake then - pushing for the development of a custom yaml export instead of relying upon the RAVEN export with potential modifications. I fully support the merging of the two.

Hao-Chalmers:
I also support the merging, which does appear to the right time point after evolvements.

BenjaSanchez:
@JonathanRob @mihai-sysbio @Hao-Chalmers sounds great! Anyone from the human-GEM / Metabolic Atlas side that would like to take up this? I'm not familiar with all these additional Metabolic Atlas-specific fields... I can imagine that a `metabolic_atlas` flag that defaults to false would be the most sensible way of doing it? I opened https://github.com/SysBioChalmers/RAVEN/pull/309 that already adds quotes to strings, so work could be continued from there?

Hao-Chalmers:
I can start to work on this, maybe also migrate `readYaml` to RAVEN, in addition to `writeYaml`.

Hao-Chalmers:
@mihai-sysbio I do not view #71 as a mistake, but improvement or enhancement instead.

@BenjaSanchez the first issue about `writeYaml` migration is how to deal with  `metadata` in RAVEN? The metadata info will add two new fields `version` and `description` to model spec and adjust existing `annotation` field.

BenjaSanchez:
@Hao-Chalmers are you referring to the 1st field in the human-GEM yml `metaData`? In practice this should not be a problem for `writeYaml`, as you could choose to write it only if the information is available in the model structure under your current format, and otherwise skip it, keeping compatibility with models that don't have that data.

Having said that, is the standard for `metaData` documented somewhere? Perhaps these missing field should become a part of `RAVEN_structure_fields.csv`? Although I do see a `description` field already matching there, what's the difference between the two?

Hao-Chalmers:
@BenjaSanchez will start a [RAVEN](https://github.com/SysBioChalmers/RAVEN) issue to address this.

mihai-sysbio:
@Hao-Chalmers which issue in RAVEN should I be watching? These are open:
- https://github.com/SysBioChalmers/RAVEN/issues/77
- https://github.com/SysBioChalmers/RAVEN/issues/107
- https://github.com/SysBioChalmers/RAVEN/issues/311

Hao-Chalmers:
@mihai-sysbio both #77 and #107 should be closed, and I did.

There is pending conclusions to #311, which seems to be relevant to `standard-GEM` as well.

Issue #174, opened 2020-06-01
feat: new sets of id for reactions and metabolites


Hao-Chalmers:
### Description of the issue:
* Here a new set of reaction id is proposed, to promote the application of `Human-GEM` as a template for generic usage
* This new set of rxn ids will be added as a new field to the annotation file `humanGEMRxnAssoc.JSON`, without affecting the model

#### Planned implementation:
* Append a new field `rxnMAID` to ~`humanGEMRxnAssoc.JSON`~`reactions.tsv`
* The element ids in `rxnMAID` will be the same as those in `rxns`, except that the prefix `HMR_` will be systematically replaced with `MAR` (abbreviation of `Metabolic Atlas reactions`)
* Append a new field `metMAID` to `metabolites.tsv`

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

JonathanRob:
Is the plan to eventually assign an `MA_` ID for non-`HMR_` reactions as well?

Hao-Chalmers:
Yes @JonathanRob, this plan aims toward assigning every `rxns` id with a unique `MA_` identifier eventually.

mihai-sysbio:
Seeing that this change has been in (slow) progress for some time, I'm wondering when it would be a good time to make this change in the entire model. I think the merging of PR #203 would make it easier to implement the adoption of new identifiers.
Another question outside the scope of this issue is if metabolites should also use `MA_` identifiers.

Hao-Chalmers:
@mihai-sysbio honestly, myself look forward to a quick adaption to entire model.

mihai-sysbio:
I can see some of the benefits of this transition, but I'm having a hard time spotting downsides besides the obvious backwards compatibility, which can be preserved by moving previous identifiers to the annotation. What other downsides are there?

Hao-Chalmers:
After discussing with @mihai-sysbio and @JonathanRob, the following consensus was reached regarding this transition:
* Expand the scope of `MA` ids from reactions to metabolites, and probably also compartments
* The `MA` id is formatted as concatenation of two parts: prefix letters and suffix digits
* The prefix are: `MAR` for reactions, `MAM` for metabolites, `MAC` for compartments
* The suffix digits are numbers, with fixed width, in sorted (or random?) order
Any comments to this proposed naming convention are welcome.

mihai-sysbio:
[...]

PR #175, opened 2020-06-02
feat: add `rxnMAID`field to JSON file

Hao-Chalmers:
* This PR is to add `rxnMAID` field to `humanGEMRxnAssoc.JSON` file, as discussed in #174


JonathanRob:
Two comments/suggestions:

1. Maybe leave the `rxnMAID` field blank for reactions that do not have an ID of the form `MA_####`, instead of just using their original ID? I realize it's temporary and probably doesn't matter, but this would at least keep all of the IDs in the `rxnMAID` field consistent.

2. Consider adding an extra digit, since we're already surpassing 10,000 reactions; e.g., `MA_3905` would be `MA_03905`.

Hao-Chalmers:
@JonathanRob the elements are reformatted as suggested.

PR #178, opened 2020-06-07
chore: sync JSON file content with model

Hao-Chalmers:
* Remove `HMR_4362 ` and associated external ids from rxn JSON file to sync with the model


JonathanRob:
Ah, good catch, @Hao-Chalmers. We should set up another automatic test to verify that the model is synced with the JSON annotation files, so this doesn't happen again.

Hao-Chalmers:
@JonathanRob will add such a test function later.

PR #187, opened 2020-07-15
refactor: use Python 3, has latest pip for cobra

mihai-sysbio:
This PR migrates Python to version 3, which removes the need to upgrade `pip` before installing `cobra`. It also addresses the `FutureWarning` message which cobrapy outputs.


mihai-sysbio:
Great improvement @Hao-Chalmers 👍 for the action to pass it needs the model file in the new location, ie #188 should be merged first.

JonathanRob:
Agreed, nice addition, @Hao-Chalmers.

A small question - should this new `basicCheck.py` script be placed in the `.github/workflows/` directory?

Hao-Chalmers:
@JonathanRob not sure about the location. @mihai-sysbio what do you think?

Some thoughts about `basicCheck`: now it has checks for cobrapy import and consistency with annotation files. More tests may be added later either by curator or user, to whom the `code/test` folder probably is better to work with?

mihai-sysbio:
Imho tests should be designed to run on their own, and just called in the workflow, so that users can have the freedom of testing whenever they prefer; I'd leave tests in `code/test`.

JonathanRob:
Ok, sounds good @mihai-sysbio @Hao-Chalmers! No need to move things around then.

PR #188, opened 2020-07-15
Align repository with standard-GEM template

JonathanRob:
This PR aims to adopt the [standard-GEM](https://github.com/MetabolicAtlas/standard-GEM) repository template for genome-scale metabolic models.

Specifically, the `.standard-GEM.md` checklist was retrieved from the [standard-GEM](https://github.com/MetabolicAtlas/standard-GEM) repository, and completed to reflect the Human-GEM repo's current observation of the guidelines.


mihai-sysbio:
I'd like to highlight a section from `standard-GEM`:
> Releases must use the tag format `X.X.X` where X are numbers, according to [semantic versioning principles](https://semver.org/). The last field (“patch”) can also be used to indicate changes to the repository that do not actually change the GEM itself. The use of a `v` before the version number (`v1.0`) is [discouraged](https://semver.org/#is-v123-a-semantic-version).

For Metabolic Atlas I've changed release tag from with `v` to without `v`. I'm am _not_ suggesting to do the same, just making sure you're aware and welcoming your opinion on the topic.

JonathanRob:
> My suggestion would be to add a data/README.md

@mihai-sysbio Yes, and as I started looking into this, I realized that the `data/` directory could *really* use some cleaning up. I think a large amount of the content there should be removed, since it is no longer used or maintained. And on that note, we could also soon delete many of the old functions in the `code/modelCuration/` subdirectory, since that is also quite cluttered and largely deprecated.

I don't think it's worth writing a README for the `data/` directory until this cleaning is done. So maybe we can implement that in another PR. What do you think?

mihai-sysbio:
Sounds like a good plan to me! Perhaps it's easier to start from an issue with a checklist.

Hao-Chalmers:
@mihai-sysbio would be good to start with a checklist.

@JonathanRob yes, the `data` folder accumulated quite a few outdated files that need to be cleaned. Regarding functions/scripts under `code/modelCuration/` subfolder, I think they should mostly kept if actually used before.

JonathanRob:
@Hao-Chalmers I would still recommend that we remove the older model curation scripts (the more recent ones can stay). They will always remain in the git history, so we will not lose any information. Leaving them there clutters the repository without any added benefit, in my opinion.

Hao-Chalmers:
@JonathanRob yes the old scripts do appear messy now. We may restructure the folder by moving old ones to designed subfolder and stop further curation.

Considerations about why to preserve the previous scripts:
- From the perspective of user, it would be important to quickly trace back to the code for making specific changes;
- From the perspective of curator, it would be really convenient to be able to identify errors directly from old scripts;
- From the perspective of repo, it's trivial amount of space for storing the code.

JonathanRob:
@Hao-Chalmers 
> From the perspective of user, it would be important to quickly trace back to the code for making specific changes;
From the perspective of curator, it would be really convenient to be able to identify errors directly from old scripts;

These can both be done very easily on GitHub or any git client.

> From the perspective of repo, it's trivial amount of space for storing the code.

I'm not concerned about space. It's a matter of keeping things clean and easy to find. The more old scripts we keep, the more difficult it is to find what you're actually looking for.

mihai-sysbio:
I believe what @Hao-Chalmers is ultimately after is improving the ease of use. While I agree fully with your perspective @JonathanRob, I think the needs Hao brings up could be addressed as follows:

- for users, create a new section in `code/README.md` or directly in the main `README.md` about _Curation history_ , explaining that curation is done through code, but that inevitably code gets out of date, therefore one has to go back in time to a specific model version to see the repository at that point in time `git checkout v1.3.1`. Alternatively, to see differences between versions, one should search in the pull requests, or from the command line run commands like `git diff v1.3.0 v1.3.1`.

- for curators, I would suggest trying `git bisect` - it saved me a couple of times. Perhaps this [tutorial](https://www.youtube.com/watch?v=D7JJnLFOn4A) is a good place to start.

- for the repo: imho, the repo has to serve the users. I find the mix of code that works and code that doesn't to be confusing, especially after taking long breaks (or for new users). Therefore, I tend to remove code that's not intended to be used in the future.

Hao-Chalmers:
@mihai-sysbio yes the repo ultimately serves for users who usually cannot be predefined.

Let's merge this PR first and find a solution via #190.

PR #195, opened 2020-09-08
feat: add new functions

Hao-Chalmers:
* Add new functions for expanding the applications of Human-GEM
	* `getModelFromOrthology` constructs a draft GEM based on a template model and provided ortholog pairs
	* The `extractAllianceGenomeOrthologs` function extracts ortholog pairs from the JSON files downloaded from Alliance Genome database and prepare for GEM generation
	* `replaceGrRules` is able to replace `grRules`, `rxnGeneMat ` and `genes` fields in a model with another set of gene IDs


code/getModelFromOrthology.m
```diff
@@ -0,0 +1,76 @@
+function [draftModel, removedRxns] = getModelFromOrthology(templateModel,orthologPairs)
```
JonathanRob, line 1: the `removedRxns` output is never actually assigned/used.

JonathanRob (approved):
Will be useful to have these functions.

I hope it's ok that I made a few minor edits related to comments and code refactoring, it shouldn't change any actual functionality.

Hao-Chalmers:
@JonathanRob Very good changes!

Just one question to [36b0ddd](https://github.com/SysBioChalmers/Human-GEM/pull/195/commits/36b0ddd99833cc965229cc9c66d03181cd97236a), the purpose was to get rid of `rxnFrom` and `metFrom` fields that are non-standard. It seems they are retained after `setdiff`?

JonathanRob:
@Hao-Chalmers That was a mistake, I meant to use `intersect`, not `setdiff`. I'll Correct that now.

Issue #200, opened 2020-09-21
fix: retired Ensembl gene ids


Hao-Chalmers:
### Description of the issue:
* The gene ids used in current Human-GEM is from Ensembl version 95, in which there are two gene ids have been retired in the latest Ensembl release v101:
    * [ENSG00000281028](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000281028) was replaced with [ENSG00000038210](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000038210)
    * [ENSG00000285441](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000285441) was replaced with [ENSG00000112096](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000112096)

#### Expected feature/value/output:
* Update the id mapping information according to the latest Ensembl version
* Update the two retired gene ids with their successors

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

PR #201, opened 2020-09-25
fix: remove retired genes

Hao-Chalmers:
This PR is to address the issue raised in #200
* Ensembl id mapping file was updated from version 95 to 101
* the genes `ENSG00000281028` and `ENSG00000285441` were abandoned in Ensembl, and merged into `ENSG00000038210` and `ENSG00000112096`, respectively. Since both replacement genes had already been included in Human-GEM, here `ENSG00000281028` and `ENSG00000285441` were removed from the model


JonathanRob:
It looks like the large diff associated with updating the Ensembl association file (a43d3cf) might be due to a re-ordering of the rows - I think the file was originally sorted by Ensembl ID. I wonder if it is possible to sort again to try and reduce the number of line changes? Or maybe it's not a big enough issue to be worth the time.

Hao-Chalmers:
It's possible to reorder the rows in another PR, so leave this PR open. Was the file originally ordered by `Gene_stable_ID`?

JonathanRob:
You mean reorder them in another commit within this PR?

Yes, they are ordered by `Gene_stable_ID`. However, `Gene_stable_ID` has a ton of duplicates, so sorting by that column alone will not give a unique ordering. It looks like the original file was sorted by the first column (`Gene_stable_ID`), then the second (`Transcript_stable_ID`), then third, etc.

Hao-Chalmers:
Okay. Will try to update with a re-ordered file in another PR, see if it works better.

Hao-Chalmers:
@JonathanRob one Ensembl association file sorted by the columns was added (commit [ea866b6](https://github.com/SysBioChalmers/Human-GEM/commit/ea866b6a2e56a204391d53eb74f839eff82a9d58)) to a new [branch](https://github.com/SysBioChalmers/Human-GEM/tree/issue200New). This re-ordered file introduced much less diff changes (`17,698` additions  and `4,311` deletions) than before (`113,114` additions and `99,727` deletions). So I think we can use this sorted version by cherry picking the other two commits from the old branch to this new one and making a new PR. The old PR can then be deprecated.

Hao-Chalmers:
The issue #200 has been resolved by another PR #202, so this branch will be deleted.

PR #202, opened 2020-09-26
fix: remove retired genes (new)

Hao-Chalmers:
This PR is to address the issue raised in #200
* Ensembl id mapping file was updated from version 95 to 101
* the genes `ENSG00000281028` and `ENSG00000285441` were abandoned in Ensembl, and merged into `ENSG00000038210` and `ENSG00000112096`, respectively. Since both replacement genes had already been included in Human-GEM, here `ENSG00000281028` and `ENSG00000285441` were removed from the model

- [X] Passed the test with `testYamlConversion` function


JonathanRob (approved):
Great work, @Hao-Chalmers!

Issue #157, opened 2020-02-05
Annotation file for genes


pecholleyc:
### Description of the issue:

The repo offers associations/annotations for reactions (`humanGEMMetAssoc.JSON`) and metabolites (`humanGEMRxnAssoc.JSON`) but there is no such json file for genes. Some gene's external identifiers can be found in the `ensembl_ID_mapping.tsv` file but this is a different format, and I think the list of Ensembl IDs in that file is not only bounds to the Human-GEM model.

#### Expected feature/value/output:

It would be useful to have a `humanGEMGeneAssoc.JSON` file that can be parsed similarly to the other association files.

#### Current feature/value/output:

The file does not exist.

#### Reproducing these results:

Yes

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

link to #75

Hao-Chalmers:
Sounds like a valid request.

Hao-Chalmers:
A new annotation file `humanGEMGeneAssoc.JSON` could be added with following fields to begin with:
`Ensembl Gene ID`
`UniProt SwissProt ID`	
`Gene symbol`
`NCBI Entrez ID`

JonathanRob:
I guess we would want to format the gene annotation structure in the same way that we have done for the reaction and metabolite annotation files. Therefore, I suggest we have a `genes` field that is identical to the `genes` field in the model structure (just as we had a `rxns` and `mets` field in the reaction and metabolite files, respectively). We could still keep the `Ensembl Gene ID` field, even though it would be the same as the `genes` field.

Also, to maintain consistency with the other annotation structures and with COBRA, here are my suggestions for the field names:
`geneEnsemblID` (or `geneENSGID`)
`geneUniProtID`
`geneNames`
`geneEntrezID`

The reason I'm leaning toward `geneENSGID` instead of `geneEnsemblID` is that it makes it easier to also add `geneENSPID` and `geneENSTID` fields if we want.

JonathanRob:
This issue was resolved in #204.

pecholleyc:
Hi, so this issue was reopened for, I guess, 2 main reasons that should be discussed in details:
1. Provide additional annotations for genes in the repository.
[...]

Issue #203, opened 2020-09-30
More accessible format of model annotations


JonathanRob:
### Description of the issue:

Currently, the model annotation information (e.g., reaction and metabolite KEGG IDs, metabolite HMDB IDs, etc.) is stored in the SBML version (.xml) of the model, or the annotation `.JSON` files: [humanGEMMetAssoc.JSON](https://github.com/SysBioChalmers/Human-GEM/blob/master/data/annotation/humanGEMMetAssoc.JSON) and [humanGEMRxnAssoc.JSON](https://github.com/SysBioChalmers/Human-GEM/blob/master/data/annotation/humanGEMRxnAssoc.JSON).

Although the information can be retrieved by loading the JSON files into Matlab, the JSON format does not seem to be the most intuitive or human-readable format for such information. I think it would be convenient to instead store this data in a tabular format (e.g., `.tsv`) that can easily be read in its raw form, imported by other languages (python, R), or viewed in Excel, for example. Furthermore, if a change is made to a reaction or metabolite (or a metabolite is deleted/added), it would be nice if it all ends up on a single line, rather than being spread across multiple lines. 

I don't recall what the original justification of using the `.JSON` format was, so please let me know if I'm missing a major advantage. Also, I don't think [standard-GEM](https://github.com/MetabolicAtlas/standard-GEM) yet advises on how to format such information.

My suggestion would be to change the `.JSON` files to a `.tsv` or similar format. However, I know that this would likely require reworking some pipelines, namely with importing the data into [Metabolic Atlas](https://www.metabolicatlas.org/). There is also an issue (#157) suggesting that we convert the gene annotation file from `.tsv` to `.JSON` - I fully support having these all in one location/format, but why not `.tsv`?

Another option (or in addition to the one above) could be to enhance the export of the [Excel model version](https://github.com/SysBioChalmers/Human-GEM/blob/master/model/Human-GEM.xlsx) so that all the annotation information is appended as additional columns in that document. That would mean that we continue using the same `.JSON` annotation files, but users could instead extract the data from the `.xlsx` file (though I am generally hesitant to trust Excel not to silently change content, such as gene names to dates).

Please let me know what your thoughts are on this; in particular @Hao-Chalmers since you established the JSONs, and @pecholleyc since this will affect import into Metabolic Atlas, and @mihai-sysbio if anything in the context of `standard-GEM` relates to this issue.

#### Expected feature/value/output:
A more accessible/human-readable annotation file format.

[...]

PR #204, opened 2020-10-02
feat: provide annotation file for genes

Hao-Chalmers:
This PR is to address the issue discussed in #157
* a `JSON` file with gene annotation info is provided by extracting relevant associations from `ensemble_ID_mapping.tsv`
* the `basicCheck` script was adapted by adding the consistency check for genes, and renamed to `sanityCheck`


mihai-sysbio (commented):
I understand the intention was to not approach #203 in this PR. My personal approach would have been to skip #157 and solve #203 directly.
Regardless, I'm wondering why does the annotation files need to be checked routinely, rather than them being re-created with every new release.

Hao-Chalmers:
> I'm wondering why does the annotation files need to be checked routinely, rather than them being re-created with every new release.

Recreating `tsv` format annotation files with every release might be a good solution to #203

JonathanRob:
@mihai-sysbio

> Regardless, I'm wondering why does the annotation files need to be checked routinely, rather than them being re-created with every new release.

I'm not sure I understand how one would "re-create" these files, since the information in those files is not stored anywhere else (except the gene annotations).

@Hao-Chalmers 

> Recreating tsv format annotation files with every release might be a good solution to #203.

Just to clarify, are you suggesting that tsv format files are generated _in addition to_ the JSON files, or that they will _replace_ the JSON files?

Also a very minor point, but why not just name the script based on what it is doing? i.e., `consistencyCheck`.

Hao-Chalmers:
Since this PR aims to address #157, I'll post responses to `tsv` annotation files under #203 for better tracking of the information flow.

mihai-sysbio:
@JonathanRob regarding the re-creation of these annotation files: somebody somehow generates them (which by the way is not really transparent how), and then they are routinely checked for consistency. Instead of that, I'm suggesting to automate the generation of these files, and to thus skip the consistency check. This doesn't need to be implemented here; #203 would be a good place in my opinion.

mihai-sysbio (approved):
A tiny comment: the branch name doesn't follow any particular convention.

JonathanRob:
@mihai-sysbio I think that's the problem - there is no way to "re-generate" these files, as far as I know. They were originally created by taking information that was stored in the model, and moving it to a different format (`JSON`) (i.e., deleted from the model). These annotation JSON files were then curated over time in parallel with the model. So, as far as I know, these JSON files are the **only** source of this information, and therefore we cannot re-generate them from some other source.

However, this could partially change in the future, if we decide to source some of the annotations from other databases, such as MetaNetX.

Hao-Chalmers:
@mihai-sysbio regarding the branch, which is often to address particular issues listed, if it's a good idea to directly name it by the issue number?

code/test/sanityCheck.py
```diff
@@ -12,20 +12,10 @@ def load_yml(yml_file):
...
-        modelMets.append(m.id)
-
-    # get list of genes
-    modelGenes = []
-    for g in model.genes:
-        modelGenes.append(g.id)
+    # get lists of rxns, mets and genes
+    modelRxns  = list(map(lambda element : element.id, model.reactions))
+    modelMets  = list(map(lambda element : element.id, model.metabolites))
+    modelGenes = list(map(lambda element : element.id, model.genes))
```
mihai-sysbio, line 18: Nice! How about reusing the _lambda_ function like so:
> fetchId = lambda x : x.id
extractIds = lambda componentArray : list(map(fetchId, componentArray))
[ modelRxns, modelMets, modelGenes] = list(map(extractIds , [model.reactions, model.metabolites, model.genes] ))

Disclaimer: the above might need some testing.

mihai-sysbio:
@Hao-Chalmers In MetAtlas we tend to follow the same style as for the semantic commits. In this particular case, I would have probably gone for something like `feat/gene-annotation_157`.

code/test/sanityCheck.py
Hao-Chalmers: this idea was actually considered, but seems against simplicity and readability rules.

mihai-sysbio: I see. The first bit though (`fetchId = lambda x : x.id`) should increase readability, unless you expect some variation in the future in how the id will be stored across different component types.

PR #208, opened 2020-10-19
feat: improve badges in README.md

### LINKED ISSUE #194: feat: tag Human-GEM releases to Zenodo
### Description of the issue:
* According to the [Zenodo discussion](https://github.com/MetabolicAtlas/standard-GEM/issues/14) on `standard-GEM`, a number of advantages could be taken through tagging the model releases to Zenodo. This feature therefore should be included to Human-GEM.

* [x] Added a DOI badge that always points to the latest release in Zenodo 
* [x] Update `README.md` file to adapt to this change

- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

mihai-sysbio:
The aim in this PR is to fix #194. At the same time, by adding the `version.txt` file, the version badge will be fixed from `?.?.?` to the right version, once this file ends up on the `master` branch.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -31,5 +31,6 @@ jobs:
 
     - name: Test import with cobrapy and consistency with annotation files
       run: |
+        pip install -U pip
         pip install cobra
         python code/test/sanityCheck.py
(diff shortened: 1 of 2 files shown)


Hao-Chalmers:
Not sure what's wrong with the version badge, which was certainly alright in this morning.

mihai-sysbio:
Oh that's a bit confusing, `version.txt` exists on `master` but not on `devel`.

Hao-Chalmers:
Now the version badge is back to normal.

mihai-sysbio:
So, what about that `version.txt` file then? Should it or should it not be present on `develop`?

Hao-Chalmers:
"files should not be added without necessity." - law of Ockham's razor

JonathanRob:
From what I remember, the problem we encountered with also having the version number/file on the `devel` branch was that it would become outdated unless we merged `master` into `devel`, which I don't think is consistent with the current workflow.

mihai-sysbio:
To resolve this PR, I propose that I remove the file `version.txt` for now, and maybe we can have a discussion about the workflow to see if there is a way to simplify it.

Hao-Chalmers:
Agree to remove the `version.txt` that can only be existed in master branch.

mihai-sysbio:
The `version.txt` file has been removed - it will not be added to `devel` in this PR.

Hao-Chalmers (approved):

PR #210, opened 2020-10-20
feat: add new function

Hao-Chalmers:
* Upload function `addMetabolicNetwork` for integrating new reactions and metabolites into a model


JonathanRob (approved):
Looks good! Should be useful when incorporating new groups of reactions from other sources/databases.

PR #211, opened 2020-10-21
fix: add glpk as dependency for swiglpk

mihai-sysbio:
There seems to be a problem in recent runs of the Actions. `swiglpk` doesn't get installed because of a missing dependency `glpsol`. By adding that dependency, we can hopefully get the workflow to work again.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -29,6 +29,9 @@ jobs:
       with:
         python-version: '3.x'
 
+    - name: Install GLPK, dependecy for swiglpk
+      run: sudo apt-get install glpk-utils libglpk-dev
+
     - name: Test import with cobrapy and consistency with annotation files
       run: |
         pip install cobra


Hao-Chalmers:
it worked

Hao-Chalmers (approved):

Hao-Chalmers:
now this PR is blocked because the only commit eff6ee7 in this branch was found from another PR

mihai-sysbio:
I would have suggested to merge this first though, in order to update the other branches that need it.

Hao-Chalmers:
yes, thought so too. but why did you cherry pick this commit elsewhere?

mihai-sysbio:
Instead of cherry-picking, I simply merged in the branch with the fix. This does not lead to any conflicts.

Hao-Chalmers:
if they were merged, we should simply deal with #208

mihai-sysbio:
I think that PR has a different purpose, as it opens a discussion into the workflow. I would propose that I undo that branch merge there and merge this current PR first.

Hao-Chalmers:
ok

Issue #189, opened 2020-07-17
Empty MIRIAM cells in Excel file


pecholleyc:
### Description of the issue:
As I understand it, the external identifiers / MIRIAM fields have been removed from the Matlab model, and are stored in JSON files now. The function to export the model to Excel format uses a RAVEN function that still expect to read the fields from the model. Consequently all MIRIAM columns in the exported Excel file are now empty.

#### Expected feature/value/output:
Add back the MIRIAM data to the Excel sheets. One way to do it would be to re-populate a temporary Matlab model object with the MIRIAM data before exporting into Excel.

#### Current feature/value/output:
Empty cells in MIRIAM columns in the Excel file

By comparing the Yeast-GEM Excel file with the Human-GEM one I found additional differences:
 - no SHORT NAME for genes in Human
 - reaction REFERENCE are only PMID numbers e.g. '21623372; 7765825' in Yeast but are prefixed and not necessary PMID in Human (e.g. 'PMID:367367;PMID:4203909;UNIPROT:P00558')
 - looks like values are separated with '; '  in Yeast but ';' in Human

So it is possible to have both Excel files formatted the same way? Following the standards if any. I am asking this because since Yeast do not use JSON for annotations, I am considering using the Excel to extract such information.

**I hereby confirm that I have:**
- [x] Tested my code on my own computer for running the model
- [x] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

mihai-sysbio:
Looking forward to see what others think. In the meantime, here's a random idea: it might be possible to extract annotation from an SBtab file, which can be obtained by using a [converter from SBML](https://www.sbtab.net/sbtab/default/converter.html).

JonathanRob:
Export of the SBML version has the same issue, in that RAVEN draws annotation information from the MIRIAM fields. This [is addressed](https://github.com/SysBioChalmers/Human-GEM/blob/7e5212625f1068b6f4519114c83d8b393ce4444f/code/io/exportHumanGEM.m#L109) by calling the `annotateModel` function prior to exporting the SBML version.

A quick fix would be to call the `annotateModel` function prior to the .xlsx export as well. This actually should have been done in the first place, but I missed it.

As for the differing fields, I'm not sure if there are standards defined somewhere, but we can work together to migrate to a common format.

JonathanRob:
[...]

PR #212, opened 2020-10-23
Feat/new annot format issue203

JonathanRob:
Addresses issue #203 by re-formatting the reaction, metabolite, and gene annotation `.json` files in the `data/annotations/` subdirectory as `.tsv` files. The files names were also simplified a bit:

| Before | After |
| ------ | ------|
| humanGEMRxnAssoc.JSON | reactions.tsv |
| humanGEMMetAssoc.JSON | metabolites.tsv |
| humanGEMGeneAssoc.JSON | genes.tsv |

A few advantages to note:
- More human-readable format
- Changes to the same reaction, metabolite, or gene are now on the same line
- File sizes are reduced by ~30%

Since MATLAB's delimited file I/O functions are, in my opinion, a garbage fire, I needed to develop some functions to simplify the import and export of these (and other) `.tsv` files: `importTsvFile` and `exportTsvFile`, respectively.

Finally, annotation data that is expected to be text/strings is enclosed in double quotes ("") in the `.tsv` files, whereas numeric data does not use quotes. Not only was this necessary to automate the `tsv` I/O process, but I figured it was also safer to explicitly define the datatype of each column, so there is no guessing involved.

#### Contributor guidelines
The contributor guidelines were updated to better reflect our current workflow.


mihai-sysbio:
Random idea: would it make sense to store these under the `model/` folder instead of `data/annotations/` ? I'm thinking they would be more obvious to external people looking into fetching the model.

JonathanRob:
@mihai-sysbio That might actually be a good idea, I was also thinking that these annotations are somewhat "buried" in the repository.

Hao-Chalmers:
@mihai-sysbio sounds good, but how could we get opinions from external users?

mihai-sysbio:
@Hao-Chalmers I've asked for feedback on [Gitter](https://gitter.im/SysBioChalmers/Human-GEM#), so we can give it a few days.

Hao-Chalmers:
very good

Hao-Chalmers (approved):
This branch nicely addressed the issue #203 - converting annotation file format: `json` -> `tsv`.

In addition, it also attempted to solve the bug reported in #189, plus update **Contributor guidelines** to reflect the latest repo status.

Since the branch already included too many commits/changes, I think it's good to go now.

JonathanRob:
Thanks, @Hao-Chalmers. We can implement the change of annotation file location (in `model/` instead of `data/annotations/`) suggested by @mihai-sysbio in a separate PR.

Issue #149, opened 2019-11-25
MOXD1 potentially incorrectly associated with HMR_4740


JonathanRob:
### Description of the issue:
A Metabolic Atlas user noted that the `HMR_4740` reaction:

`1-piperideine-6-carboxylate[c] + H+[c] + H2O[c] <=> 2-aminoadipate 6-semialdehyde[c]`

is associated with the MOXD1 gene, despite the gene encoding a monooxygenase with no known substrate. Furthermore, the references associated with the reaction did not contain any evidence of such an association.

After some investigation, it was found that this gene-reaction association originated from the EHMN model, yet no evidence supporting MOXD1 involvement is provided. Therefore, it is likely that this gene is incorrectly associated with the reaction and should be removed, but it should first be investigated further.

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

JonathanRob:
I still have not been able to find any evidence supporting the involvement of MOXD1 in this reaction (HMR_4740).

Interestingly, there also doesn't seem to be evidence for the other gene associated with this reaction, DBH. In fact, this reaction is reported as spontaneous ([KEGG](https://www.kegg.jp/entry/R02317), [MetaCyc](https://biocyc.org/META/NEW-IMAGE?type=REACTION&object=RXN-8173)), suggesting that this reaction should be associated with neither MOXD1 nor DBH.

I therefore plan to delete the `grRule` (`MOXD1 or DBH`) for reaction `HMR_4740`.

JonathanRob:
Gene association removed in #215.

PR #215, opened 2020-12-06
Fix: Issue 149

JonathanRob:
Addresses issue #149 by deleting the GPR associated with reaction [`HMR_4740`](https://metabolicatlas.org/explore/gem-browser/human1/reaction/HMR_4740). The described functions of the genes currently associated with the reaction (MOXD1 and DBH) do not suggest that they would catalyze this reaction, and reaction databases ([KEGG](https://www.genome.jp/dbget-bin/www_bget?rn:R02317), [MetaCyc](https://biocyc.org/META/NEW-IMAGE?type=REACTION&object=RXN-8173), [Rhea](https://www.rhea-db.org/rhea/25124)) annotate this reaction as "spontaneous". Therefore, the reaction is now annotated in Human-GEM as spontaneous, and its gene associations were removed.

Since MOXD1 is not associated with any other Human-GEM reactions, it was removed entirely from the model.


Hao-Chalmers (approved):

Hao-Chalmers:
@JonathanRob this seem good to go

Issue #214, opened 2020-12-05
mistaken GPR annotation of pyruvate dehydrogenase 


XuhangLi:
### Description of the issue:
It seems the GPR annotation of HMR_4137 (pyruvate dehydrogenase) is mistaken. According to the GPR annotation of Recon3D (http://bigg.ucsd.edu/models/Recon3D/reactions/PDHm) and literature search (i.e,PMID: 22675509), the PDHA2 gene functions as an isoenzyme of PDHA1. Therefore, the GPR should be revised to DLD and PDHX and (PDHA1 or PDHA2) and DLAT and PDHB.

The current GPR resulted in false predictions in the application of gene expression integration. 
#### Expected feature/value/output:
DLD and PDHX and (PDHA1 or PDHA2) and DLAT and PDHB

#### Current feature/value/output:
DLD and PDHX and PDHA1 and DLAT and PDHA2 and PDHB

#### Reproducing these results:
NA

**I hereby confirm that I have:**
- [x] Tested my code on my own computer for running the model
- [x] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

*Note: replace [ ] with [X] to check the box. PLEASE DELETE THIS LINE*

Hao-Chalmers:
@XuhangLi thank you very much for reporting this!

This GEM repository aims to serve, and also welcome contributions from, the research community.

The `Human-GEM` applies a `yaml-based` curation workflow that can be easily implemented by any participants. You can conduct the proposed changes by following steps:
1. fork `Human-GEM` repo to your local 
2. first switch to `devel` branch, and then start a new fixing branch
3. make the GPR change and update the reference directly to the `Human-GEM.yml` file within the fixing branch
4. generate a pull request for merging your fixing branch to the `devel` branch of `Human-GEM`

Thank you again and look forward to your pull request.

JonathanRob:
Thank you for identifying this, @XuhangLi! I agree that the GPR annotation you have proposed looks correct.

To add to @Hao-Chalmers's comment, just let us know if you would prefer that we implement this fix for you.

XuhangLi:
Hi Hao and Jonathan, 

Thanks your timely response! I am glad that I could help with the human model development. 

Per @JonathanRob's response, does it mean you guys will do the pull request for me? It would be great that I don't need to learn the yaml-based workflow, but I would be happy to do so if it is needed.

Hao-Chalmers:
[...]

PR #216, opened 2020-12-07
fix: GPR for HMR_4137

mihai-sysbio:

This PR addresses #214.


Hao-Chalmers:
@mihai-sysbio thanks for starting this fixing PR as proposed in #214.

To assist efficient machine parsing and convenient knockout simulation, another format is suggested as below according to RAVEN issue [#69](https://github.com/SysBioChalmers/RAVEN/issues/69)

Current:
`"ENSG00000091140 and ENSG00000110435 and (ENSG00000131828 or ENSG00000163114) and ENSG00000150768 and ENSG00000168291"`

Suggested:
`"(ENSG00000091140 and ENSG00000110435 and ENSG00000131828 and ENSG00000150768 and ENSG00000168291) or (ENSG00000091140 and ENSG00000110435 and ENSG00000163114 and ENSG00000150768 and ENSG00000168291)"`

mihai-sysbio:
That's a pretty long issue to read and understand @Hao-Chalmers - could you share a link to the specific message that proposes the solution? Perhaps some clear/easy documentation can be added to Human-GEM.

Hao-Chalmers:
The suggested `grRules` format can be found at the end of this [function](https://github.com/SysBioChalmers/RAVEN/blob/master/core/standardizeGrRules.m).

JonathanRob:
@mihai-sysbio To summarize, the issue arises due to the way that RAVEN (specifically the `removeGenes` function) works with gene rules.

Currently, it recognizes groups of isozymes - i.e., genes or complexes separated by `or`. Therefore, it will break the following rule:

`geneA or (geneB and geneC)`

into:

[`geneA`, `(geneB and geneC)`]

It will then process the pieces separately. However, this algorithm was not built to deal with enzyme complexes containing isozymic subunits. For example, the following rule:

`geneA and (geneB or geneC)`

will be processed into:

[`geneA and (geneB`, `geneC)`]

Notice that the rule "pieces" now contain unbalanced parentheses, which is bad. Therefore, RAVEN has enforced a "standard" rule format, where formats like the second example are not allowed. Instead, they recommend that the second rule is instead re-written as:

`(geneA and geneB) or (geneA or geneC)`

Logically, it is identical, but unfortunately it can end up being very long for some rules.

mihai-sysbio:
Thanks for the details @JonathanRob!

I'm persisting in finding a nicer solution to this. To me, it is unbearable that a user, who is typically not an expert in modelling, has to follow an undocumented and complicated rule because of a peculiarity in a tool (RAVEN) that is only available in one programming language (Matlab) that is also not even the most popular tool available. On top of this, neither RAVEN nor Matlab are required to be used when curating the yaml file, which is a key point of the yaml-based workflow. 
In my view, when importing the yaml file with Matlab, a few functions could be added so that the rules are automatically converted to the desired format. Here's a sketch based on the example above:
```
>> syms x y z;
>> expr = x & (y | z);
>> E = expand(expr);
>> E
 
E =
 
x & y | x & z
```

Before expanding, some other functions could be useful, eg
```
S = simplify(expr, 'steps', 50);
```

If it works, perhaps this pre-processing of the rules could be useful in GECKO too.

JonathanRob:
I'm very glad you mentioned this, @mihai-sysbio - I developed a `simplifyGrRules` function a while back that does this (if the `expanded` input is set to true).

Unfortunately, there are a few details to consider before implementing it automatically:
1. Some gene rules become extremely long when expanded, and MATLAB will error when attempting to expand these. Even if it does successfully expand the rule, the length can make it much less human-readable.
2. I believe there are still some gene rules with gene associations that would be lost if the expression was simplified. For example: `G1 or (G1 and G2)` simplifies to just `G1`, so the G2 association with that reaction would be lost.

Both of these problems could be addressed by identifying the problematic/affected reactions (probably not too many), and attempting to revise the gene rules so they no longer suffer from these issues. Maybe an issue should be created about this.

mihai-sysbio:
That's interesting.

What I wanted to propose is to allow rules to be curated in the way that makes most biological sense. Then, either as the last step of importing into Matlab, or as a first step of running GECKO to run `simplifyGrRules` followed by `expand`.

The issue that remains is 
> Some gene rules become extremely long when expanded, and MATLAB will error when attempting to expand these. Even if it does successfully expand the rule, the length can make it much less human-readable.

In my opinion, the expanded rules should never be stored in the source `yaml`, to keep them both human-readable and biologically relevant.

So, what to do when Matlab errors on expanding rules? Well, that's not a `yaml` problem, nor a Raven problem. I'd think it would be worth digging deeper into solving this within Matlab.

JonathanRob:
If you're interested, there are only 4 problematic grRules in Human-GEM: `HMR_9491`, `HMR_9577`, `HMR_9579`, and `ATPasel`. The first three fail to convert to a symbolic expression because they are "too complex" (too many genes):

```
Error using str2symInternal (line 91)
Unable to convert string to symbolic expression:
L 1 (C 1): MXASET: Code analysis did not complete. File contains expressions that are too complex.
```

Interestingly, I looked at these first three grRules and they are already in standard/expanded form; so they're actually ok.

The fourth grRule is more difficult. It can be converted to a symbolic expression, but becomes too large when expanded and eventually you run into the error above. Maybe we can manually curate this rule to get around this issue.

Hao-Chalmers:
@JonathanRob good work, which helps to clarify the status of `grRules` formats in Human-GEM.

The GPR [rule](https://metabolicatlas.org/explore/gem-browser/human1/reaction/ATPasel) of `ATPasel` looks overwhelmed and requires manual investigations. This probably can be addressed in another issue/PR.

Here I suggest we just focus on `HMR_4137` and update its `grRules` with an standard/expanded form.

Hao-Chalmers (approved):

Hao-Chalmers:
let's just move on

XuhangLi (commented):
the GPR modification looks good, but the references seem a little unrelated. I suggest to include an additional reference that supports PDHA2 being an isoenzyme of PDHA1 (so we use "or" connections). PMID: 2249846 could be a good choice for this.

Hao-Chalmers:
@JonathanRob I'm think could `simplifyGrRules` be more practically useful by taking one or multiple GPR rules as input, instead of the entire `grRules` field?

JonathanRob:
The input to `simplifyGrRules` just needs to be a cell array, so it doesn't have to be the entire `grRules` field. It will work with just one or a few rules.

Hao-Chalmers:
sounds good

PR #218, opened 2020-12-14
feat: new function for gap-filling

Hao-Chalmers:
* a gap-filling function, `gapfill4EssentialTasks`, was developed for introducing reactions from the reference model into an input model, so that the resulting model can undertake all essential tasks, including growth on Ham's media. The changes introduced by this function are recorded in `modelChanges` variable for manual inspection. By default, the biomass objective equation is set to "biomass_components" that is constituted by generic components occurring in an eukaryotic cell.

### DIFF
--- a/code/gapfill4EssentialTasks.m
+++ b/code/gapfill4EssentialTasks.m
@@ -49,6 +49,17 @@ essentialTasks = fullfile(path,'../data/metabolicTasks','metabolicTasks_Essentia
 taskStruct = parseTaskList(essentialTasks);
 %taskStruct = taskStruct(end);
 
+% re-organize biomass in human reference, first block all biomass equations
+ind = find(startsWith(refModel.rxns,'biomass'));
+refModel.ub(ind) = 0;
+refModel.lb(ind) = 0;
+refModel.c(ind)  = 0;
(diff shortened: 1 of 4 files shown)


JonathanRob (approved):
Very nice addition. I tested this on a few examples by randomly deleting a subset of reactions, and it worked fine.

code/gapfill4EssentialTasks.m
```diff
@@ -102,6 +102,8 @@
 modelChanges = docModelChanges(model_orig, outModel);
 
 
+end
```
JonathanRob, line 105: This doesn't change anything, but I believe that sub-functions should come _after_ the main function `end` statement, so I have corrected that here.

Hao-Chalmers: Now this sub-function just blocks the other biomass reactions. I wonder if it's better to just delete them?

JonathanRob: I don't think it makes a difference, so maybe just leave it as is for now.

Hao-Chalmers: ok

Issue #190, opened 2020-07-20
style: whether to preserve previously used curation scripts or not?


Hao-Chalmers:
### Description of the issue:
* There have been accumulated many curation scripts in the repo. These scripts were used only once for making specific changes to the model, and won't be used again. This issue is to bring up discussions about whether to preserver these one-time scripts under `code` folder or not?

#### Expected feature/value/output:
* Find a way that serves ease of use for both users and curators

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

mihai-sysbio:
Based on the reasoning in #188, I think @JonathanRob summarized it well:
> remove the older model curation scripts (the more recent ones can stay)

Together with this, it would be nice to have a guide on how to go back in time and walk through an example curation.

JonathanRob:
Yes, I like @mihai-sysbio's suggestion to provide a short guide on how to take advantage of the git framework. I guess adding such a guide to the README in the `code/` directory would be best, as you suggested earlier, rather than the main repository README. Another option would be to create a wiki for the repository and add it there, but maybe that's an idea for the more distant future.

Hao-Chalmers:
@mihai-sysbio @JonathanRob I don't have strong opinions to the removal of old scripts, but I do concern that this solution might be biased toward curators, instead of users.

JonathanRob:
To add to this, I would also recommend removing a lot of the content in the `data/` directory. Much of the data and scripts in that directory are no longer used or maintained, so they should be removed to avoid confusion/clutter.

Hao-Chalmers:
Agree, the outdated content under `data`  and useless scripts in `code` can be gradually moved a root directory `deprecated`, in which no further curation will be made and can be completely removed later at some time point (e.g. major release).

mihai-sysbio:
The move to a `./deprecated` folder might further impede the ability to run said code because of broken folder structure. Moreover, the recovery mechanism is the same - going back in time to a previous release. Therefore I find the move to such a folder unnecessary.

Hao-Chalmers:
1. The move to a `deprecated` folder is to stop running and curating the code/data any further;
2. If rely on the recovery mechanism along, may all the content of `data` and `code` be just deleted?
3. Many data files and code appear to be unused and not necessary, but if this is 100% certain?
4. If it's a big cost to have such a folder with ~100Mb in size?

[...]

Issue #176, opened 2020-06-06
style/doc: cleanup authorship and timestamps


mihai-sysbio:
### Description of the issue:
I've seen the practice of including author names at the top of code files in `Human-GEM`. I'd like to open a discussion around this.
I feel author names at the top of a file can look prohibitive. But more importantly, I don't think this practice serves its purpose that well. To figure out who authored a file with git one can use [git blame](https://www.git-scm.com/docs/git-blame) or even easier [GitHub Blame](https://help.github.com/en/github/managing-files-in-a-repository/tracking-changes-in-a-file).
Perhaps this would also be a good time to reconsider the need for lines such as:
> % DATE CREATED:
%      MODIFIED:

#### Expected feature/value/output:
Removal of authorship and timestamp lines.

#### Current feature/value/output:
Logging of authorship and timestamp in code files.

#### Reproducing these results:
```
grep -nr "keyword" *  
```
where possible keywords are: Hao, Jonathan, DATE

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [ ] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

Hao-Chalmers:
Thanks @mihai-sysbio for bring up this issue that could be relevant to all GEM-type repos.

Hao-Chalmers:
Would be nice to have a broader participation to this discussion @JonathanRob @simas232 @edkerk @BenjaSanchez

JonathanRob:
> where possible keywords are: Hao, Jonathan, DATE
Hah! I like the idea of my name being used as a keyword to find lines of code to remove.

But yes, I think this system is outdated and I support the removal of such lines.

BenjaSanchez:
I also don't think it makes sense to keep that info, I usually Github blame anyways.

mihai-sysbio:
If authorship stats are ever needed, the [Contributors page](https://github.com/SysBioChalmers/Human-GEM/graphs/contributors) allows for time interval selection, but is limited to the master branch. One can also install `git-extras` and run [git summary](https://github.com/tj/git-extras/blob/master/Commands.md#git-summary) on any branch. That being said, the repository is quite noisy in the number of lines so I would other metrics.

edkerk:
Agree to remove authorships and timestamps, just have a good look at the diff to make sure that not too many lines are deleted.

Hao-Chalmers:
[...]

PR #223, opened 2021-01-15
Move deprecated data and code to new .deprecated directory

JonathanRob:
Moves all deprecated files, scripts, and models in `data/` that are no longer maintained or in use to a `.deprecated` folder in the main directory. The purpose of this is to avoid clutter and potential confusion by keeping these files, and is discussed in #190. The deleted files can always be accessed by retrieving an earlier version of this repository, so the information is not lost and can be restored if we determine that anything was wrongly deleted.

A `README` file is now also added for the `data/` directory to be consistent with [standard-GEM](https://github.com/MetabolicAtlas/standard-GEM) guidelines.


JonathanRob:
One file that I was uncertain about was `currencyMets.tsv` in the `data/Maps/` subdirectory. I wasn't able to determine what this is/was used for, so let me know if it should **not** be deleted and I will restore it.

mihai-sysbio:
I really like the chosen name `.deprecated` 👍🏻

Hao-Chalmers:
The `currencyMets.tsv` was used in Metabolic Atlas development for map generation, during which the metabolites included in this file were omitted in 2D maps for clarity.

mihai-sysbio:
It might make sense to move `currencyMets.tsv` to  the [Human-maps repository](https://github.com/sysbiochalmers/human-maps), if this file is not relevant for modelling.

pecholleyc:
Not sure it should be moved to that repo, this file was used a one given time, on Human-GEM maps freshly generated from HMR 2.0 maps, but then the currency state of metabolites was manually/semi-auto set for each reactions, without using the information in the file. I think it was also used during db generation for Metabolite Atlas, when the currency state was displayed on the front-end.

mihai-sysbio:
@pecholleyc thanks for the background info.
Metabolic Atlas will handle the currency state differently (same currency state for all models), so it will not use this file for that purpose. This file is present in previous `Human-GEM` releases, and since it's not meant to be moved to the `Human-maps` repo, I'm in support of not keeping it around.

Hao-Chalmers (approved):

README.md
```diff
@@ -108,8 +108,8 @@ Contributions are always welcome! Please read the [contribution guidelines](http
...
-- [Pınar Kocabaş](https://www.chalmers.se/en/staff/Pages/kocabas.aspx), Chalmers University of Technology, Gothenburg Sweden
-- [Pierre-Etienne Cholley](https://www.chalmers.se/en/staff/Pages/cholley.aspx), Chalmers University of Technology, Gothenburg Sweden
-- [Avlant Nilsson](https://www.chalmers.se/en/staff/Pages/avlant-nilsson.aspx), Chalmers University of Technology, Gothenburg Sweden
-- [Hao Wang](https://www.chalmers.se/en/staff/Pages/hao-wang.aspx), Chalmers University of Technology, Gothenburg Sweden
+- [Jonathan L. Robinson](https://bii.dk/team/jonathan-robinson/), The BioInnovation Institute (BII), Copenhagen, Denmark
+- [Pınar Kocabaş](https://research.chalmers.se/en/person/kocabas), Chalmers University of Technology, Gothenburg, Sweden
+- [Pierre-Etienne Cholley](https://www.chalmers.se/en/staff/Pages/cholley.aspx), Chalmers University of Technology, Gothenburg, Sweden
+- [Avlant Nilsson](https://www.chalmers.se/en/staff/Pages/avlant-nilsson.aspx), Chalmers University of Technology, Gothenburg, Sweden
+- [Hao Wang](https://www.chalmers.se/en/staff/Pages/hao-wang.aspx), Chalmers University of Technology, Gothenburg, Sweden
```
mihai-sysbio, line 115: I hope it's okay to bring this up again about maintaining a _Contributors_ section https://github.com/SysBioChalmers/Human-GEM/issues/176#issuecomment-640478034

mihai-sysbio (approved):
I've had a quick look but couldn't find the place where this `.deprecated` folder is introduced. Perhaps it would make sense to mention it somewhere in the Readme.

README.md
JonathanRob: It wasn't clear to me from the linked thread, but are you suggesting to just remove this contributors section from the Readme because it's redundant with information already present on GitHub?

mihai-sysbio: >  remove this contributors section from the Readme

Yes that was what I meant. Now the _Contributors_ are even highlighted on the right part of the landing page of the repository. Also, Yeast-GEM has already removed this section.

JonathanRob: Ah ok, then yes, I completely agree.

mihai-sysbio:
This looks great!

PR #240, opened 2021-03-11
[FIX] changed pubchem ID for m01778c

jorgemlferreira:
Fixed pubchem ID for m01778c


Hao-Chalmers:
In addition to PubChem, should other metabolite ids (e.g. KEGG, MetaNetX) be modified as well? @jorgemlferreira

jorgemlferreira:
Hello again @Hao-Chalmers. So I've made a few changes that are present in the changelog provided on the commit. I think this part may be debatable, since most of the IDs provided by the model led to the conjugate base form of the metabolite. I don't know if this is the kind of identifiers you want in your model, so it's up for debate, since I believe most of the fatty acids are treated in the same manner. However, the initial change to the pubchem IDs are needed, and to be changed in all the compartments (I didn't do it in my first commit).

Please let me know if you just want changed the pubchem IDs, I'll rollback to the original .tsv and change only that.

Best regards

Hao-Chalmers:
@jorgemlferreira the changes you made look okay based on the provided `Changelog`. But we'll need more time to evaluate them, and will come back to you later.

jorgemlferreira:
@Hao-Chalmers Hi again. No problem. I'll assume this kind of changes could lead to a high amount of work, and may not even be necessary, depending on how you want the model to work (specially in fatty acids metabolism).

Best

Hao-Chalmers:
@jorgemlferreira thanks for the input with detail changelog and accurate InChI mapping! Really surprised by that the commit message could be so long, and can be a good way of log changes.

The modifications to PubChem and CHEBI ids are good. I would suggest to keep the ids from HMDB and LipidMaps though. Because CHEBI and PubChem have very detail classification, while the others don’t. The HMDB and LipidMaps ids are acceptable and good for inclusiveness. In this sense, the KEGG id for `ocdcea` should also be retained. What do you think?

jorgemlferreira:
@Hao-Chalmers Hello again, and thank you for the kind words. I agree with you, mainly due to how the fatty acids metabolism is built in the model. However, at some point in the future (maybe when another significant breakthrough is done in this area) it should be considered to revisit and rebuild the whole fatty acids metabolism, as a community effort.

I agree with only changing the PubChem and CHEBI ids, since the others are not 100% incorrect. I'll rollback the changes and only change the ids mentioned before.

Best,

Jorge Ferreira

JonathanRob:
@jorgemlferreira I am also very much in support of a complete redesign of the fatty acid metabolic network portion of Human-GEM, though I realize this is no small task.

jorgemlferreira:
@JonathanRob I think everyone wins in this scenario. You've made an important step in making this model and allowing the community to contribute in every way they can. And yes, I know this is not a simple and/or easy task, but a collaboration between the main and small developers/users of human metabolic models could make an integrated effort to improve this, and this repository is an important tool for that. Let's hope this can happen, and if it happens, I'll gladly help in anything I can.

Best,

Jorge

Hao-Chalmers (approved):

Issue #224, opened 2021-01-21
feat: Improve reaction, metabolite, and gene annotations


JonathanRob:
### Description of the issue:
Human-GEM reactions, metabolites, and genes are associated to many external identifiers from databases such as KEGG, BiGG, ChEBI, and MetaNetX. However, many associations are outdated or missing. It would be great to not only update old IDs, add missing IDs, and correct any erroneous IDs, but to do so in an automated manner such that we can simply re-run the pipeline again in the future to ensure that the IDs remain current.

#### Expected feature/value/output:
An automated or semi-automated pipeline/script that updates old/missing/incorrect reaction, metabolite, and gene identifiers present in the `reactions.tsv`, `metabolites.tsv`, and `genes.tsv` annotation files. It may be beneficial to develop this in python so it can be implemented as a GitHub action.

#### Current feature/value/output:
Many model components are missing annotation information and/or may have outdated or incorrect annotation information.

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

Hao-Chalmers:
This is certainly an important issue.

However, it appears that there is no readily acceesbile pipeline(s) that are tuned for GEM curation. We thus welcome the contributions and collaborations from, as well as serve for, the research community toward such a (semi-)automated pipeline.

mihai-sysbio:
>  update old IDs, add missing IDs, and correct any erroneous IDs

Intuitively, these look like different problems requiring independent solutions. Based on how the curation progress so far, it would look to me like adding missing IDs has do be done manually. A script (GH Action) could be set up to cross-reference IDs to highlight the erroneous ones, but resolving the conflicts manually might be more straightforward, at least in the near future.

> re-run the pipeline again in the future to ensure that the IDs remain current

Once there is an ID that gets outdated perhaps a new one can be provided by the reference database.

jorgemlferreira:
Hello. Don't know if this is the right way to do this, but since in this thread is about annotations, both m01778c (elaidate) and m02646c (oleic acid, oleate) share the same pubchem.compound ID (445639), whereas m01778c should be 5461071 and m02646c should be 445639. If it is necessary to open a new thread, I'll do it. I'd be a good idea to open an issue where people could report this kind of issues when dectected, and be integrated into the pipeline.

Best regards

Hao-Chalmers:
@jorgemlferreira please open an issue for this, and very welcome for the contribution!

jorgemlferreira:
[...]

PR #241, opened 2021-03-19
Rhea annotations

pecholleyc:
 A total of **590** Human-GEM reactions were associated Rhea ID, and the corresponding master Rhea id. More information about the mapping process to Rhea can be found [here](https://github.com/SysBioChalmers/Human-GEM/issues/224#issuecomment-797778447)
- 69 were mapped via gene id -> UniProt id -> MetaNetX rxn id -> Rhea id
- 487 were mapped via equation mapping
- 34 were mapped by both methods


model/reactions.tsv
```diff
@@ -6024,11 +6024,11 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "HMR_5984"	""	""	""	"r2069"	""	"r2069"	"MNXR105978"	"HMR_5984"	"RCR40526"	"2.A.3.8.15"	0	"MA-R05984"	""	""
 "HMR_5985"	""	""	""	"r2070"	""	"r2070"	"MNXR105979"	"HMR_5985"	"RCR40160"	"2.A.3.8.15"	0	"MA-R05985"	""	""
 "HMR_5986"	""	""	""	"r2071"	""	"r2071"	"MNXR105980"	"HMR_5986"	"RCR41449"	"2.A.3.8.15"	0	"MA-R05986"	""	""
-"HMR_5987"	""	"FE2t"	""	"r2072"	""	"FE2DMT1"	"MNXR99505"	"HMR_5987"	"RCR40527"	"2.A.55.2.3"	0	"MA-R05987"	""	""
+"HMR_5987"	""	"FE2t"	""	"r2072"	""	"FE2DMT1"	"MNXR99505"	"HMR_5987"	"RCR40527"	"2.A.55.2.3"	0	"MA-R05987"	"RHEA:29441;RHEA:29440;RHEA:29581;RHEA:29580"	"RHEA:29439;RHEA:29579"
```
pecholleyc: Here, the Rhea master IDs can be difficult to associate programmatically to the Rhea IDs (2 master IDs versus to 4 Rhea IDs). Do you think the content should be changed, for instance by always having the same number of IDs in the two columns (this implies to duplicate Rhea master IDs)?

model/reactions.tsv
```diff
@@ -8426,7 +8426,7 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "FAS80COA_L"	""	"FAS80COA_L"	""	""	""	"FAS80COA_L"	"MNXR99415"	""	""	""	0	"MA-R01036"	""	""
 "FBA5"	""	"FBA5"	""	""	""	"FBA5"	"MNXR99463"	""	""	""	0	"MA-R01038"	""	""
 "FCLTm"	""	"FCLTm"	""	""	""	"FCLTm"	"MNXR99471"	""	""	""	0	"MA-R01044"	"RHEA:22584"	"RHEA:22584"
-"FE2tm"	""	"FE2tm"	""	""	""	"FE2tm"	"MNXR99501"	""	""	""	0	"MA-R01046"	""	""
+"FE2tm"	""	"FE2tm"	""	""	""	"FE2tm"	"MNXR99501"	""	""	""	0	"MA-R01046"	"RHEA:29441;RHEA:29440;RHEA:29581;RHEA:29580"	"RHEA:29439;RHEA:29579"
```
pecholleyc: Same problem for that reaction.

model/reactions.tsv
pecholleyc: Another solution would be to manually curate the association now, some Rhea ID can be probably removed.

Hao-Chalmers: Among the four Rhea ids mapped to `HMR_5987`, only `RHEA:29581` appears to be the correct association while the others can be ruled out according the reaction direction or metabolite compartment.

Hao-Chalmers:
The most recent commit introduced some new associations that include one-to-multiple links between **Human-GEM** and **Rhea**. This is completely fine if the links can be verified programmatically or manually. Otherwise, they probably should be leave out. 

From the point of view of reviewer, it's alright to have less associations as long as they are accurate.

mihai-sysbio:
One-to-many links increase the difficulty of using the annotations as they need to be treated differently in code. If possible, sticking to a single id simplifies the work.

Hao-Chalmers:
@pecholleyc it's pleasant to go through these Rhea associations, especially the approach via equation mapping that nicely link to the ones with proper reaction direction. While the way through UniProt link to Rhea ids in undefined directions. Could the latter be revised with proper direction by using the equation mapping method?

Hao-Chalmers (approved):

Hao-Chalmers:
@pecholleyc were these Rhea IDs with directed reaction updated manually, without the help of equation mapping?

PR #245, opened 2021-04-12
feat: add MA met ids to metabolites.tsv

Hao-Chalmers:
* a new column of `metMAID` is added to `metabolites.tsv`, according to the plan proposed in #174:
    - The MA met ids of non-HMR metabolites are formatted as `MA-MDDDDDC`, where the prefix are in the range of MA-M00001 to MA-M09999 and appended with a compartment id. They are ensured in an ordered and unique filling.
    - For metabolites inherited from HMR, their MA met ids are obtained by simply adding "MA-" prefix:

| mets          |  metMAID     |
| ------------- | --------- |
| "m01967c" | "MA-M01967c" |


mihai-sysbio (approved):
nice!

code/modelCuration/README.md
```diff
@@ -1,6 +1,5 @@
 # Model curation scripts
 
-This directory contains curation-related scripts and functions used to make changes to the Human-GEM model. These model curation scripts help to improve transparency of changes made to the model when the number of changes is too large to view practically.
-
-Note that all code in this directory is considered deprecated and will not be updated or maintained.
+This directory contains curation-related scripts and functions used to make changes to the Human-GEM repository. These curation scripts help to improve transparency of changes made to the model when the number of changes is too large to view practically.
 
+- `addMAMetIDs.m` and `completeMARxnids.m` are scripts for adding metabolite and reaction identifiers (#174) that are to be used for Met Atlas web-portal.
```
JonathanRob, line 5: We don't typically include descriptions of individual functions or scripts in README files; is there a reason you chose to do so in this case? I would recommend writing the description/purpose at the top of the scripts themselves, rather than separating the description from the file.

Hao-Chalmers: done

Issue #250, opened 2021-04-19
Discontinued/replaced NCBI gene id


haowang-bioinfo:
### Description of the issue:
* It is found that in the gene annotation file `genes.tsv` one associated Entrez gene id [**107986084**](https://www.ncbi.nlm.nih.gov/gene/?term=107986084) was discontinued by NCBI and replaced with **7375**. 

#### Expected feature/value/output:
* The discontinued/replaced NCBI ids can be updated to the latest ones
* Ideally this update could be automated by the `get_ensembl_ID_mapping.py` script mentioned in #157

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

JonathanRob:
I noticed that the updated `genes.tsv` file (on `devel` after merging #251) contains the following entry in the NCBI (`geneEntrezID`) column for this gene: `7375;107986084`. I'm assuming that with later Ensembl updates that the deprecated ID (107986084) will be removed, but at the moment it's partially fixed.

Since the updated workflow with `genes.tsv` does not allow for manual modification of this file (thereby offloading the maintenance responsibilities to Ensembl), I will close this issue under the assumption that it will soon be fixed on Ensembl.

Hao-Chalmers:
@JonathanRob no this issue is not fixed, let's keep it open.

JonathanRob:
@Hao-Chalmers my reason for closing this issue is because, with our current workflow, the responsibility of maintaining the proper gene ID associations is now placed on Ensembl. So in reality, the issue should be made with Ensembl, not Human-GEM, since we do not have the power to fix this. But you can keep the issue open if you want to make users aware of the temporary existence of an expired NCBI ID in our `genes.tsv` file.

pecholleyc:
But I agree to close this issue. Moreover I think the order of the IDs in `genes.tsv` is not random, Ensembl always puts the more meaningful ID at the left most position.

Hao-Chalmers:
> Ensembl always puts the more meaningful ID at the left most position.

@pecholley if you have any evidence supporting this?

mihai-sysbio:
@haowang-bioinfo I remember briefly looking in the documentation for the answer (evidence), but couldn't find it. Since both @JonathanRob and @pecholleyc were happy to close the issue, I will go ahead and close it.

PR #244, opened 2021-03-31
feat: more annotation for genes

mihai-sysbio:
### Main improvements in this PR:
Metabolic Atlas had enhanced the annotation of genes outside the Human-GEM. With this PR, the aim is to migrate the annotation closer to the model. This can also be seen as a contribution to #224. Once this is merged, the original file `GENES.tsv` from the repository [MetabolicAtlas/data-files](https://github.com/MetabolicAtlas/data-files) may be removed.

**I hereby confirm that I have:**

- [ ] Tested my code on my own computer for running the model
- [x] Selected `develop` as a target branch

PR #251, opened 2021-04-21
feat: update gene annotation script

pecholleyc:
Improve the python script used to generate the ID mapping table (`ensembl_ID_mapping.tsv`):

- the script writes the `genes.tsv` file directly.
- the column names have changed following the decisions in issue #157
- the two new columns of PR  #244 are generated and included in the output file
- The script can now be imported as a module and be used in other python script
- The newly generated `gene.tsv` fixes issue #250

The idea is also to generate the `genes.tsv` file when the sanityCheck.py complains on the genes IDs comparison. A PR that updates the `sanityCheck.py` will be created, once this PR has been accepted.


code/GPRs/fetch_ensembl_gene_annotations.py
```diff
@@ -0,0 +1,266 @@
...
+import requests
+import re
+from collections import OrderedDict
+
+"""
+Fetch and write in OUTPUT FILE gene annotations of the input YAML file. The file must contains Ensembl IDs as gene IDs.
+Input:
+ - the YAML file format of the GEM
+ - version of Ensembl, e.g. 103
+ - version of Ensembl, e.g. 103
```
mihai-sysbio, line 14: This line appears duplicated
```suggestion
```

mihai-sysbio:
Really nice stuff @pecholleyc! Random thought: would it be too much work to generalize the script for other (non-human) models?

pecholleyc:
There is maybe some additional works in Matlab script because of the removal of the `ensembl_ID_mapping.tsv` :neutral_face:

JonathanRob:
> There is maybe some additional works in Matlab script because of the removal of the `ensembl_ID_mapping.tsv` 😐

Yep, I'm working on these changes and will update them soon.

pecholleyc:
The script is already 'kind of' generalized for non-human model, there is a `--database` argument that can be used to target any model in Ensembl. But I never tested with another DB, I hope the main SQL queries is still valid. The problem will be maybe with the query that fetches the primary assembly IDs; It can be tricky to generalize because of the name of the chromosomes are different between species.

Hao-Chalmers:
> There is maybe some additional works in Matlab scripts because of the removal of the ensembl_ID_mapping.tsv 😐

a random thought: since a new script "fetch_ensembl_gene_annotations.py" was introduced here, the removal of `get_ensembl_ID_mapping.py` and `ensembl_ID_mapping.tsv`, as well as the derived adjustments, are not necessarily a part of this branch.

pecholleyc:
@Hao-Chalmers, The new script supplant the old one and the previous logic, I don't think both should exist at the same time in the repository. I think it will make things more complicated.

JonathanRob:
@Hao-Chalmers Also, I've already started the work on this branch, it would be easier to just continue it here.

Hao-Chalmers:
> @Hao-Chalmers, The new script supplant the old one and the previous logic, I don't think both should exist at the same time in the repository. I think it will make things more complicated.

@pecholleyc true, they should not both exist in the `master` branch. While the `develop` suppose to be dynamic and with tolerance.

> @Hao-Chalmers, The new script supplant the old one and the previous logic, I don't think both should exist at the same time in the repository. I think it will make things more complicated.

@JonathanRob then go ahead

nanjiangshu:
In the YAML file `Human-GEM.yml`, `version` is empty and the term `organism` is missing. Is that done by purpose or?

mihai-sysbio:
> In the YAML file `Human-GEM.yml`, `version` is empty and the term `organism` is missing. Is that done by purpose or?

The `version` is blank on `develop` and filled in on the main branch.

nanjiangshu:
OK, for the `organism`, it can be inferred from the newly added term `taxonomy: "9606"`

Hao-Chalmers:
@nanjiangshu which branch are you talking about?

JonathanRob:
Ok, I made a few changes:

1. fd94216: Updated the `translateGrRules` function so that it uses the `genes.tsv` annotation file instead of the now-deprecated `ensembl_ID_mapping.tsv` file 

2. 7c2d4f0: Updated the `annotateGEM` function to be compatible with the new `genes.tsv` file, and to retrieve more of the fields (columns) in the other annotation tsv files.

3. 983382d: I noticed that the Rhea IDs that were recently added to the `reactions.tsv` file had the format `RHEA:#####`. This is inconsistent with how we typically store the IDs in the annotation TSVs (i.e., without the prefix). I trimmed the `REAH:` prefixes from the identifiers in the `reactions.tsv` file.

JonathanRob (approved):

mihai-sysbio:
> is is inconsistent with how we typically store the IDs in the annotation TSVs (i.e., without the prefix).

Is there any exception besides the ChEBI prefix in `metabolites.tsv` (which has to stay there)?

nanjiangshu:
> @nanjiangshu which branch are you talking about?

Hi @Hao-Chalmers , thanks for your reply. I mean the branch of this pull request, i.e., `feat/update_gene_annotation_script`.
@mihai-sysbio has alreadly answered my question in the post above. https://github.com/SysBioChalmers/Human-GEM/pull/251#issuecomment-823956404

JonathanRob:
> Is there any exception besides the ChEBI prefix in `metabolites.tsv` (which has to stay there)?

@mihai-sysbio as far as I know, ChEBI is the only weird case that doesn't follow the same url pattern as the other identifiers.

Hao-Chalmers:
looks good to me

pecholleyc:
I don't fully approve the removal of the prefix of Rhea IDs because, similar to ChEBI IDs, it is part of the identifier. I obtain them by parsing Rhea and retrieved the values provided by the `accession` key. This format is, I believe, used by other databases providing mapping to Rhea (e.g. UniProt). The problem is that Rhea is just using ##### as `id` internally (in their database), and also in their URLs. Maybe this is probably the reason it has been registered on Identifiers.org to accept only digits as ID (which is inaccurate in my opinion).

JonathanRob:
@pecholleyc Ok, I see. It seems the ID formatting is more complicated than I thought, and is not consistent among different sources.

In that case, we should undo my removal of the "RHEA:" prefix, and I can incorporate the ID adjustment into the `annotateGEM` function, as I did with the ChEBI ID.

mihai-sysbio:
@pecholleyc I think you're right, see the screenshot below when trying to match a KEGG reaction id on the Rhea-db website:

image

Even Rhea-db presents its identifiers with the `RHEA:` prefix. It's very much in line with what ChEBI does, so @JonathanRob's suggestion makes sense to me.

PR #256, opened 2021-05-16
feat: new function for updating animal GEM

Hao-Chalmers:
* add new functions `updateAnimalGEM `and `updateAnimalAnnotations` for automated updating of animal GEMs
* refactor `getSubNetworksInComp` according to user request
* ~~introduce pseudo gene **TBD** (stands for to-be-determined) as the GPRs of gap-filled reactions~~ adjust `gapfill4EssentialTasks` to the using of MA ids


mihai-sysbio:
@Hao-Chalmers is the
> automated updating of animal GEMs

also meant to handle the annotation files?

Hao-Chalmers:
> also meant to handle the annotation files?

@mihai-sysbio no, not yet.

Hao-Chalmers:
@mihai-sysbio just added a new function `updateAnimalAnnotations` that handles the annotations.

code/gapfill4EssentialTasks.m
```diff
@@ -82,13 +82,19 @@
...
 % clean the grRules associated with gap-filling reactions
 gapfilledRxns = setdiff(outModel.rxns, model.rxns);
 [~, ind] = ismember(gapfilledRxns, outModel.rxns);
 outModel.grRules(ind) = {''};
 
+% Assign "TBD" gene to the grRules which previously were non-empty in the
+% reference model. This gene name is abbreviated from "to be determined"
+[~, ind2] = ismember(gapfilledRxns, refModel.rxns);
+nonEmptyInd = getNonEmptyList(refModel.grRules(ind2));
+outModel.grRules(ind(nonEmptyInd)) = {'TBD'};
```
JonathanRob, line 96: I would avoid using `'TBD'` or any other unexpected (non-gene symbol or ID) value in the `grRules` field, since it could cause unforeseen problems. I would recommend just leaving them empty, and if you want to indicate that these reactions were added due to gap filling, then it can be noted in some other annotation field, such as rxnNotes or something similar.

mihai-sysbio:
Is it at all possible to set up the `unique` parts of the animal GEM fully compatible with `Human-GEM`? If yes, I would imagine the `rxnsToAdd` and `metsToAdd` to be described in a `.yml` file that would be combined with the `ortholog.yml`. Similar to this, the final annotation files for reactions and metabolites would result from combining two fully-compatible `tsv` files.

code/gapfill4EssentialTasks.m
Hao-Chalmers: The GPR `TBD` (stands for `to-be-determined`) was adopted from iCEL1314 model. And the purpose is separating these gap-filled reactions from enzymatic reactions (with regular GPRs) and the other spontaneous or exchange reactions that have empty GPRs.

JonathanRob: Then this information should be placed in the `spontaneous` field. Even though the iCEL1314 model used the `grRules` field for this information, it is not compatible with the way that Human-GEM, RAVEN, and COBRA have defined the content of the `grRules` field.

Hao-Chalmers:
> Is it at all possible to set up the unique parts of the animal GEM fully compatible with Human-GEM?

What do you mean by "compatible"?

code/gapfill4EssentialTasks.m
Hao-Chalmers: Why this `TBD` (or gap-filling) information should be placed in the `spontaneous` field?

JonathanRob: `grRules` is for gene-reaction rules, *not* for gap-filling notes and other information. A user (or function) could very easily confuse "TBD" as a gene symbol. I suggested the `spontaneous` field because you mentioned that the use of "TBD" was to differentiate these reactions from spontaneous reactions with empty GPRs, and the entire purpose of the `spontaneous` field is to indicate which reactions are enzymatic or non-enzymatic.

Another option would be to create a `rxnNotes` field or similar, where such information could be included.

mihai-sysbio:
> What do you mean by "compatible"?

@Hao-Chalmers by fully compatible I mean that all the `unique` components sit in the same format and data structures.

code/gapfill4EssentialTasks.m
Hao-Chalmers: > A user (or function) could very easily confuse "TBD" as a gene symbol.

This is a valid argument, and it sounds good to me by using `rxnNotes` field to store this information.

Hao-Chalmers:
@mihai-sysbio sounds a good idea. Would be good to discuss the specific implementation plan somewhere else.

code/gapfill4EssentialTasks.m
Hao-Chalmers: The inclusion of `TBD` GPR is now reverted. The gap filling information is planned to be added to `rxnNotes` field after resolving #272.

mihai-sysbio (commented):
On the surface things look alright. Perhaps some instructions on how this function is used in practice would have helped, but I'm not in the position to properly validate this.

Hao-Chalmers:
@JonathanRob gap filling information is now provided in `rxnNotes` field of added reactions.

JonathanRob (approved):

Issue #262, opened 2021-06-04
spontaneous field is quoted if it is a 1, not quoted if zero


johan-gson:
### Description of the issue:
It is just annoying when you read it in, you have to do some formatting, convert to number etc.

#### Reproducing these results:

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [ ] Done this analysis in the `master` branch of the repository
- [ ] Checked that a similar issue does not exist already

*Note: replace [ ] with [X] to check the box. PLEASE DELETE THIS LINE*

johan-gson:
just to be clear, no quotes would be preferrable

Hao-Chalmers:
@johan-gson the `spontaneous ` info was introduced by #162 as a numeric field

Hao-Chalmers:
you will get the field as double type, if using the `importTsvFile` function provided by Human-GEM.

JonathanRob:
@johan-gson is right, the file currently has quotes around `1`s. Someone must have accidentally modified this at some point.

It should be easily fixable by importing/re-exporting using the `importTsvFile` and `exportTsvFile` functions, respectively.

johan-gson:
So, I can use the proper import function, I just think it is weird, and it messes things up a bit if you want to load it in R or something.

JonathanRob:
Yes, it's not supposed to be like that, and should be fixed such that they're all numeric.

Hao-Chalmers:
Ah, that's weird.

Some considerations:
- Add introduction of `tsv` structure and manipulation to README
- Add tests to `tsv` structure in `sanityCheck`

mihai-sysbio:
With a bit of `git bisect`, it looks like the extra quoates slipped through in commit 98eb520aac6137aa013265f37f29fb490c669226 . @Hao-Chalmers good idea to also test the `tsv` files.

Hao-Chalmers:
resolved in #263

PR #263, opened 2021-06-06
fix: remove quotes from the `spontaneous` column in `reactions.tsv`

Hao-Chalmers:
* Fix bug reported in #262, by using `importTsvFile` and `exportTsvFile` functions consecutively
* Add structure information of annotation `tsv` files to README
* Add test to the type of `spontaneous` in `sanityCheck`


README.md
```diff
@@ -86,6 +81,68 @@ Additional annotation information and external identifiers for Human-GEM reactio
...
+rxnHMR2ID      |HMR2 reaction ID        |
+rxnRatconID    |Ratcon reaction ID      |
+rxnTCDBID      |TCDB ID                 |
+spontaneous    |Spontaneous status      | 
+rxnMAID        |MA reaction ID          |
+rxnRheaID      |Rhea ID                 |
+rxnRheaMasterID|Master Rhea ID          |
+
+
+* `metes.tsv` content:
```
johan-gson, line 109: metes.tsv -> mets.tsv

README.md
```diff
@@ -86,6 +81,68 @@ Additional annotation information and external identifiers for Human-GEM reactio
...
+
+filedname      |  annotation             |
```
johan-gson, line 111: is it supposed to be filedname, and not filename? Also in the table further down

README.md
Hao-Chalmers: fixed

README.md
Hao-Chalmers: it should be `fieldname `

mihai-sysbio:
@Hao-Chalmers do you know why something was modified at the end of the lines in commit 3de8d3fa29ee1de0fe3624fd89c7ff96a8954cde ? I'm thinking whatever it was, it might be a problem in the rest of the `.tsv` files as well.

README.md
```diff
@@ -102,5 +159,4 @@ A collection of manually curated 2D metabolic maps associated with Human-GEM are
 
 ## Contributing
 
-Contributions are always welcome! Please read the [contribution guidelines](https://github.com/SysBioChalmers/Human-GEM/blob/master/.github/CONTRIBUTING.md) to get started.
-
+Contributions are always welcome! Please read the [contribution guidelines](https://github.com/SysBioChalmers/Human-GEM/blob/master/.github/CONTRIBUTING.md) to g
```
mihai-sysbio: It looks like something was accidentally deleted here.

mihai-sysbio (changes requested):

code/test/sanityCheck.py
```diff
@@ -27,6 +27,8 @@ def checkRxnAnnotation(rxns):
     rxnAssoc = pd.read_table("model/reactions.tsv")
     rxnList = rxnAssoc['rxns'].to_list()
     assert rxnList == rxns, "Reaction annotation mismatch!"
+    assert pd.api.types.is_numeric_dtype(
+        rxnAssoc['spontaneous']), "Spontaneous column should be in numeric!"
```
mihai-sysbio, line 31: Good idea! If you think this can/should be done in other places as well, it would be good to track that with an issue.

Hao-Chalmers: No other places in my knowledge.

Hao-Chalmers:
> @Hao-Chalmers do you know why something was modified at the end of the lines in commit 3de8d3f ? Could you perhaps document this?

What was modified at the end of the lines?

mihai-sysbio:
> What was modified at the end of the lines?

It looks like all the lines have been modified, and a trailing character at the end of each line has been removed. It was unexpected to see it, but I guess it's good that it was removed. I wonder how it will be with the other `.tsv` files.

johan-gson:
It's not a \r by any chance? Windows generally uses \r\n instead of \n when writing files in text mode, which is a source of many problems. If that is removed is a good thing. So, if someone edits this file on Windows, that may happen I would guess.

mihai-sysbio:
@johan-gson good thinking! That might be it. I tried to see, but the editor wasn't showing these regardless of the settings for the end of line characters.

Since these changes are effectively polluting the history of the file, I would think it is better to revert these changes, so that the PR will capture only the lines that were required to change, instead of all lines in the file.

Hao-Chalmers:
> It looks like all the lines have been modified

@mihai-sysbio now I understand what you mean. It does appear to be an issue about line endings @johan-gson.

Since Windows and Unix/Linux/Mac use different newline symbols, i.e. '\r\n' and '\n', respectively. We'll run into this whenever the changes are made from different OS. The `reactions.tsv` in develop previously used '\r\n' as endings, which were changed to other type by `3de8d3fa29ee1de0fe3624fd89c7ff96a8954cde`.

Hao-Chalmers:
Well, it seems this has been taken care by git based this GitHub [document](https://docs.github.com/en/github/getting-started-with-github/getting-started-with-git/configuring-git-to-handle-line-endings).

mihai-sysbio:
> Well, it seems this has been taken care by git based on this GitHub [document](https://docs.github.com/en/github/getting-started-with-github/getting-started-with-git/configuring-git-to-handle-line-endings).

Excellent find! How about adding a [`.gitattributes` file](https://docs.github.com/en/github/getting-started-with-github/getting-started-with-git/configuring-git-to-handle-line-endings#per-repository-settings)?

Hao-Chalmers:
Sounds a good idea @mihai-sysbio.

IMHO, LF (Unix: \n) is preferred to CRLF (Windows: \r\n).

johan-gson:
Yeah, you should definitely go for \n. Good that git can solve this. If you figure out how to do it, please let me know since I have some other projects that I would like to add this to as well.

johan-gson:
There is also a linux command for converting: dos2unix I think it is called

Hao-Chalmers:
the `reactions.tsv` file is already in LF (\n) endings.

README.md
```diff
@@ -86,6 +81,68 @@ Additional annotation information and external identifiers for Human-GEM reactio
...
+rxnHMR2ID      |HMR2 reaction ID        |
+rxnRatconID    |Ratcon reaction ID      |
+rxnTCDBID      |TCDB ID                 |
+spontaneous    |Spontaneous status      | 
+rxnMAID        |MA reaction ID          |
+rxnRheaID      |Rhea ID                 |
+rxnRheaMasterID|Master Rhea ID          |
+
+
+* `mets.tsv` content:
```
JonathanRob, line 109: Note that the file name is `metabolites.tsv`, not `mets.tsv`.

Hao-Chalmers: fixed

mihai-sysbio:
@Hao-Chalmers will the `.gitattributes` be added in this PR or in another one?

Hao-Chalmers:
@mihai-sysbio another one

mihai-sysbio (approved):

johan-gson (approved):

Issue #265, opened 2021-06-11
feat: replace HMR ids with MA ids in Human-GEM


haowang-bioinfo:
### Description of the issue:
* This issue is a follow-up of #174 and aims to have a smooth transition from HMR ids to MA ids in `Human-GEM`

#### tentative implementation plan:
0. change prefix of MA reaction and metabolite ids to `MAR` and `MAM`, respectively
1. use MA reaction and metabolite ids formulated in #174 to replace `rxns` and `mets` fields in the model, respectively
2. swap the content between columns `rxns` and `rxnMAID` in `reactions.tsv`, then  rename column `rxnMAID` to `rxnRetired`
3. swap the content between columns `mets` and `metMAID` in `metabolites.tsv`, and rename column `metMAID` to `metRetired`
4. regenerate column `metNoComp` from new `mets` column

**I hereby confirm that I have:**
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

mihai-sysbio:
After some recent feedback from MetaNetX on the naming of the IDs, it has become clear that these should not contain any special characters. Therefore, the target prefixes are `MAR` and `MAM`.

Hao-Chalmers:
> After some recent feedback from MetaNetX on the naming of the IDs, it has become clear that these should not contain any special characters. Therefore, the target prefixes are MAR and MAM.

Sounds good to me.

JonathanRob:
Something to keep in mind, which is what @mihai-sysbio mentioned earlier: we need some system to keep track of which IDs have been used, so that retired IDs are not accidentally re-used later.

mihai-sysbio:
I would guess that we could use the content of the annotation files for old IDs that have been used and removed, together with expanding the check that all the current IDs should exist in the annotation files, and none of these IDs should exist in the "used" annotation files.

Hao-Chalmers:
Where to store these "used but removed" ids? @JonathanRob @mihai-sysbio

mihai-sysbio:
> Where to store these "used but removed" ids? @JonathanRob @mihai-sysbio

How about `/data`?

Hao-Chalmers:
> How about /data?

Do you mean an additional file?

mihai-sysbio:
> Do you mean having an additional file? what is the spec (filename, file type)?

[...]

PR #270, opened 2021-06-17
feat: implement MA ids for reactions and metabolites

Hao-Chalmers:
* replace `rxns` and `mets` fields in Human-GEM and annotation `tsv` files with MA ids using script `implementeMAID`, as planned in #265

- [X] Passed `testYamlConversion` check


mihai-sysbio (approved):
Looks good! Here are some thoughts, please mark as complete at your discretion:
- [ ] something should be done to the `metsNoComp` column
- [ ] the files where the "old" MA ids will be kept should be added to this PR

Hao-Chalmers:
@mihai-sysbio what do you mean by `the files where the "old" MA ids will be kept`?

Hao-Chalmers:
my thoughts:

1. the "old MA ids" will become the ids once the PR is merged;
2. in my understanding, there is no plan for for lessons 6 and 7. But it's good to have one asap;
3. yes, this PR is to address just #265

mihai-sysbio:
I see as part of #265 also to address lessons 6 and 7, and to solve these there would need to be a place to store the MA ids that get removed from the model ("old", ie used in the past but not used any longer).

Hao-Chalmers:
Agree, would like to see specific plan proposed under #265

Issue #279, opened 2021-07-19
wrong gene annotation?


PkiwiBird:
The Human-GEM model seems to have a gene (SLC22A5) that is incorrectly annotated to reaction HMR_2590. At least in the articles I have read, SLC22A5 codes for OCTN2 that transports carnitine from outside cell to inside cell (cellular uptake), whereas HMR_2590 transports carnitine between mitochondria and cytosol. If I am wrong I would kindly ask you to provide the name of the article proving otherwise.
Best Regards.

Hao-Chalmers:
@PkiwiBird thank you very much for reporting this.

You claimed that gene SLC22A5 (OCTN2) should not be assigned to reaction `HMR_2590` (now renamed to [MAR02590](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR02590)). Could you please help to clarify following questions:
1. What is the literature reference that contradicts against the above assignment?
2. In addition, should SLC22A5 be replaced with another gene? if so, what would be the new gene, and any supporting literature?
3. Should SLC22A5 be assigned to any other reaction(s) in Human-GEM?

PkiwiBird:
> @PkiwiBird thank you very much for reporting this.
> 
> You claimed that gene SLC22A5 (OCTN2) should not be assigned to reaction `HMR_2590` (now renamed to [MAR02590](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR02590)). Could you please help to clarify following questions:
> 
> 1. What is the literature reference that contradicts against the above assignment?
> 2. In addition, should SLC22A5 be replaced with another gene? if so, what would be the new gene, and any supporting literature?
> 3. Should SLC22A5 be assigned to any other reaction(s) in Human-GEM?

[...]

PR #281, opened 2021-07-19
fix: remove membrane transporter SLC22A5 from carnitine shuttle

PkiwiBird:
excluded gene ENSG00000197375 from gene reaction rule of reaction MAR02590, as discussed in #279.


Hao-Chalmers (commented):
@PkiwiBird the change looks good.

Please rename model file `Human-GEM.yml.txt` back to `Human-GEM.yml`.

mihai-sysbio:
This issue resolves #279.

mihai-sysbio:
There has been an unintentional rename that should be addressed:

Hao-Chalmers (approved):

JonathanRob (approved):
Thanks for pointing this out and fixing it, @PkiwiBird!

PR #283, opened 2021-07-21
fix: cytosolic protons

### LINKED ISSUE #282: cytosolic proton in in non-cytosolic reactions
### Description of the issue:
Reaction [MAR05259](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR05259) is pumping out a proton in cytosol, according to its reaction equation. Because of this, all of the reaction associated genes end up also being associated to the cytosol.

[MAR05259](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR05259) would be expected to have all metabolites in one compartment (lyososme) by using [MAM02039l](https://metabolicatlas.org/explore/Human-GEM/gem-browser/metabolite/MAM02039l).

#### Current feature/value/output:
MAR05259 produces [MAM02039c](https://metabolicatlas.org/explore/Human-GEM/gem-browser/metabolite/MAM02039c).

- [x] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

mihai-sysbio:
This PR resolves #282 by:
- switching the product MAM02039**c** to MAM02039**l** in reaction MAR05259
- merging MAR04758 into MAR01044

- [x] Passed `testYamlConversion` check


mihai-sysbio:
In line with [this comment in issue](https://github.com/SysBioChalmers/Human-GEM/issues/282#issuecomment-886449121): 
- MAR04758 was merged into MAR01044 via commits 5f59d42 and 492d74d
- MAR01459 was merged into MAR02041 via commit 9c7ca8b

The removal of MAR04758 and MAR01459 should also mark these IDs as deprecated, by keeping track of them. How about storing the deprecated identifiers in https://github.com/MetabolicAtlas/identifiers as discussed in #265 ?

Hao-Chalmers:
> The removal of MAR04758 and MAR01459 should also mark these IDs as deprecated, by keeping track of them. How about storing the deprecated identifiers in https://github.com/MetabolicAtlas/identifiers as discussed in #265 ?

Yes, the deprecated ids should be tracked. I suggest to begin with keeping these ones within Human-GEM under `/data`. What do you think?

mihai-sysbio:
> Yes, the deprecated ids should be tracked. I suggest to begin with keeping these ones within Human-GEM under `/data`. What do you think?

Sounds good to me - this is implemented in 919d20e.

Hao-Chalmers (approved):
looks great

mihai-sysbio:
@Hao-Chalmers thanks for updating MAR02041 in 522500e and for adding the headers in the file for deprecated reactions.

I'm just curious, where did you get the RHEA ids from, why was an EC class removed, and how come the confidence score was increased to `4`? Personally, I find these fixes not really being part of the discussion in t#282, so it's not something I would think of doing. A similar fix is updating MNXR100715, since it is deprecated and replaced by MNXR145247, but `!!! Alternative mappings exist !!!`.

Hao-Chalmers:
> I'm just curious, where did you get the RHEA ids from, why was an EC class removed, and how come the confidence score was increased to 4?

These changes came from the confirmation of the many genes associated with MAR02041, as tabulated in #282. By search these Ensembl ids in Uniprot, it was found that enzymes encoded by these genes do catalyze the conversion of estrone and with `Experimental evidence at protein level`. As a result, the confidence score was increased while the annotations (name, EC, Rhea ids) were updated.

mihai-sysbio:
Thanks for the explanation @Hao-Chalmers. I had assumed the table in #282 was purely a result of the merge, without searching in additional databases.

Hao-Chalmers:
In addition, would suggest that ChEBI and Rhea identifiers should be favored as more trusted external databases in the annotation curation of Human-GEM, to avoid potential conflict when multiple external ids are involved.

mihai-sysbio:
@FaribaRoshanzamir noticed that I have inadvertently introduced a bug in 5f59d42 by making `MAR01044` the _only_ reaction in the model with two subsystems attached to it. I think we should continue the discussion in a new issue.

Issue #269, opened 2021-06-16
redundant 'Tridecanoyl-CoA'


ANiknejad:
### Description of the issue:
similar to issue #268, Tridecanoyl-CoA is reported twice:

- M_m03050m (also M_m03050r, also M_m03050c ), aka lipidmapsM:LMFA07050397

- M_tridcoa_m, aka chebi:84061

M_tridcoa_m (chebi:84061) has the correct stereo for the CoA part and we propose you conserve M_tridcoa_m as the unique identifier for 'Tridecanoyl-CoA'.

Best regards,

Anne

**I hereby confirm that I have:**
- [ X] Checked that a similar issue does not exist already

Issue #268, opened 2021-06-16
redundant metabolite 'Undecanoyl-CoA'


ANiknejad:
### Description of the issue:
Undecanoyl-CoA, is reported twice in human_gem:

- M_m03116m (also M_m03116c)

- M_undcoa_m

M_undcoa_m is identical to chebi:77547 which has correct stereo, while M_m03116 is actually hmdb:HMDB0013114, with incorrect stereo. We propose you map and conserve M_undcoa to represent 'Undecanoyl-CoA' in your human model.

Best regards,

Anne

Anne Niknejad
SIB, Vital-IT, MetaNetX curator

**I hereby confirm that I have:**

- [ X] Checked that a similar issue does not exist already

Hao-Chalmers:
@ANiknejad thanks for reporting this.

Hao-Chalmers:
@ANiknejad are you suggesting to merge the duplicated `m03116m` and `undcoa_m` into one?

In addition to `Undecanoyl-CoA` reported here and `Tridecanoyl-CoA` reported in #269, there are other **5** similar cases as tabulated below:

| Metabolite name    |  id1   |      Name1        |      id2       |        Name2              |
|:-------------------|--------|:-------------------|-----------|:--------------------------|
| 5-Tetradecenoyl-CoA| m01141p| 5-tetradecenoyl-CoA| tetde5coa_p| 5-Tetradecenoyl Coenzyme A|
| 7-Hexadecenoyl-CoA | m01191p| 7-hexadecenoyl-CoA | hexde7coa_p| 7-Hexadecenoyl Coenzyme A |
| Heptanoyl-CoA      | m02107m| heptanoyl-CoA      | hepcoa_m   | Heptanoyl Coenzyme A      |
| Nonanoyl-CoA       | m02616m| nonanoyl-CoA       | noncoa_m   | Nonanoyl Coenzyme A       |
| Pentanoyl-CoA      | m02694m| pentanoyl-CoA      | pentcoa_m  | Pentanoyl Coenzyme A      |
| Tridecanoyl-CoA    | m03050m| tridecanoyl-CoA    | tridcoa_m  | Tridecanoyl Coenzyme A    |
| Undecanoyl-CoA     | m03116m| undecanoyl-CoA     | undcoa_m   | Undecanoyl Coenzyme A     |

Each pair appears to point to the same metabolite according to their names. Do you think should they all be merged?

JonathanRob:
[...]

PR #285, opened 2021-07-28
fix: merge duplicated long-chain fatty acyl-CoA metabolites

JonathanRob:
Issues #268 and #269 identified two long-chain fatty acyl-CoA metabolites (Undecanoyl-CoA and Tridecanoyl-CoA, respectively) that were duplicated in the model, and not identified earlier due to a difference in the "CoA" component being written out as "Coenzyme A" in the metabolite names. @Hao-Chalmers identified an additional 5 long-chain fatty acyl-CoAs that were duplicated in the same manner (see #268).

This PR removes the duplicated metabolites, and updates some of the annotation information associated with the remaining version of the metabolites.

- ~[ ] Passed `testYamlConversion` check **<-- This seems like redundant information that should be handled automatically**~


Hao-Chalmers:
quick comments:
- the check of `testYamlConversion` is to ensure lossless conversion between `yaml` and `mat` files. As a Matlab function, it cannot be conviniently incorporated into GH actions (let me know if I were wrong)
- start a new `deprecatedMetabolites.tsv` file under `/data` folder to host the rows removed from `metabolites.tsv`

JonathanRob:
Ah, ok my mistake, I thought it was referring to a python script. But yeah, if its a Matlab script then no problem.

Hao-Chalmers:
It's different from `sanityCheck`, which is to ensure the compatibility with `cobrapy` and `YAMLLint`.

mihai-sysbio:
`testYamlConversion` is now performed via GH action as a result of #286.

JonathanRob:
Fantastic, thank you @mihai-sysbio!

JonathanRob:
I have now added the deprecated metabolites (and their annotation information) to the new `deprecatedMetabolites.tsv` file.

Hao-Chalmers (approved):
very nice

Issue #278, opened 2021-07-12
Error: could not retrieve gene annotation from Ensembl


Hao-Chalmers:
### Description of the issue:
* Errors occurred when updating `genes.tsv` using script `fetch_ensembl_gene_annotations.py` with the latest Ensembl release **104** (default parameters) as below:
 
```
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000100101'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000189366'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000240224'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000243135'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000255730'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000259916'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000272916'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000284844'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000285043'
Error: could not retrieve Ensembl annotations for gene ID 'ENSG00000285269'
```

These errors could be due to format changes in annotation data (extracted by `retrieve_ensembl_gene_annotations`) and deprecated IDs ([ENSG00000240224](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000240224) was replaced with [ENSG00000288705](https://www.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000288705), [ENSG00000243135](https://www.ensembl.org/Homo_sapiens/Gene/Idhistory?g=ENSG00000243135) was replaced with [ENSG00000288702](https://www.ensembl.org/Homo_sapiens/Transcript/Idhistory?g=ENSG00000288702)). There won't be errors if specifying to release 103 (`--ensembl-version 103 --genome-version 38`).

#### Expected feature/value/output:
* There should be format changes to the annotation information of genes uniquely annotated by Ensembl, the `fetch_ensembl_gene_annotations` script therefore should be modified to adapt to these changes
* A mechanism could be potentially formulated for routine replacement of deprecated gene IDs

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

pecholley:
[...]

PR #292, opened 2021-08-05
fix: replace deprecated gene ids

Hao-Chalmers:
* Ensembl ids are used as gene identifiers in Human-GEM. As reported in #278, two ids (ENSG00000240224 and ENSG00000243135) were replaced by new ones (ENSG00000288705 and ENSG00000288702, respectively) in the latest release v104
* This PR is to sync above changes from Ensembl to Human-GEM


JonathanRob (approved):
Looks good!

Issue #44, opened 2018-10-15
The rules for deleting problematic/unused/wrong reactions in humanGEM


Hao-Chalmers:
### Description of the issue:
* Removal of reactions is commonly occurred in the curation of genome-scale models, and should be done carefully
* This issue is for proposing and developing a set of feasible rules that would help in implementing the deletion of problematic/unused/wrong reactions in **humanGEM**
* The removal of identical/similar reactions is discussed in elsewhere (PR #43), and will not be covered here

#### Expected outcome:
* Below are the proposed  **rules** for deleting problematic/unused/wrong reactions are drafted as below for discussion:
	1. Always apply soft deletion (**ub**=**lb**=0) when deleting problematic/unused/wrong reactions, so that the inactivated reactions can be conveniently restored;
	2. Add the list of inactivated reactions to a tab-delimited log file for (automatic) implementation;
	3. Hard deletion (complete removal) is allowed only when upgrading to a major release (e.g. human v1.0.0), either by script or command;
	4. Setup a defined format (e.g. in `mat`, `tsv`) for saving the physically removed reactions for tracking changes.

#### Current state:
* So far, the removal of unused reactions in **humanGEM** has happened once by [script](https://github.com/SysBioChalmers/HMR3-GEM/blob/master/ComplementaryScripts/modelCuration/miscModelCurationScript_20180921.m) for preparing the release [v0.3.1](https://github.com/SysBioChalmers/HMR3-GEM/releases/tag/v0.3.1). However, this removal was found inappropriate and quickly restored by another [script](https://github.com/SysBioChalmers/HMR3-GEM/blob/master/ComplementaryScripts/modelCuration/miscModelCurationScript_20181012.m) for the release [v0.4.2](https://github.com/SysBioChalmers/HMR3-GEM/releases/tag/v0.4.2).
* It is the above removal and restoration of reactions that induced this issue for developing proper rules for deleting problematic/unused/wrong reactions
* A soft deletion approach (i.e. constraining both upper and lower bounds to zero) has been suggested in another issue (PR #33), and started to be implemented by the [script](https://github.com/SysBioChalmers/HMR3-GEM/blob/innerMembraneComp/ComplementaryScripts/modelCuration/curateMitochMembraneComp.m) for curating the reactions relevant to inner mitochondrial compartment

**I hereby confirm that I have:**
- [x] Checked that a similar issue does not exist already

Hao-Chalmers:
[...]

PR #307, opened 2021-08-31
Fix/removed beta-D-glucose and related reactions

pranasag:
Human-GEM had two definitions of cytosolic glucose and glucose-6-phosphate: `glucose[c]` and `beta-glucose[c]` (and respectively for the G6P). The conversion rxn `beta-D-glucose[c] <=> glucose[c]` (`MAR07745`) and the reactions with the "beta-" glucose, ending with `fructose-6-phosphate[c]` create unnecessarily alternative to glycolytic pathway from `glucose[c]` to `fructose-6-phosphate[c]`, which has no added value to the model. Moreover, the "D-" and "unspecified" stereoisomer distinction is not there for any other sugar. Thus I suggest removing the respective metabolite species (`MAM01388c` and `MAM01389c`) and reactions (`MAR07745`,`MAR07746`, `MAR07747`,` MAR07748`, `MAR07749`).


Hao-Chalmers:
@pranasag interesting propose.

> The conversion rxn beta-D-glucose[c] <=> glucose[c] (MAR07745) and the reactions with the "beta-" glucose, ending with fructose-6-phosphate[c] create unnecessarily alternative to glycolytic pathway from glucose[c] to fructose-6-phosphate[c], which has no added value to the model. 

May I know what evidence make these reactions unnecessary to the model?

pranasag:
Hi @Hao-Chalmers, 
First, for all the reactions with `beta-D-glucose[c]` or `beta-D-glucose-6-phosphate[c]`, there are concurrent reactions, using `glucose[c]` or `glucose-6-phosphate[c]`, so the pathway remains there (and maintains a classical shape of glycolysis).
Second, I agree with your observation that mutarotases, such as GALM, are detected to be expressed in experiments and have a meaning there, because glucose in solution is in an equilibrium between the alpha- and beta rotational isomers (not sure whether this influences kinetic parameters of enzymes, though). However, neither mutarotation is relevant for reaction stoichiometry nor is defined completely in the model (then we'd have also a set of reactions for alpha-D-glucose as such, perhaps?). That'd be my rationale for removing this set of reactions.

By the way, I'm not sure why the YAML validation failed (says - reaction annotation not matching), I checked the changes introduced and these should be only related to removing 2 species and 5 reactions, as suggested.

Hao-Chalmers:
@pranasag thanks for quick response.

> By the way, I'm not sure why the YAML validation failed (says - reaction annotation not matching), I checked the changes introduced and these should be only related to removing 2 species and 5 reactions, as suggested.

The failure of YAML validation is due to the inconsistency between model (in YAML) and annotation files ([`reactions.tsv`](https://github.com/SysBioChalmers/Human-GEM/blob/main/model/reactions.tsv) and [`metabolites.tsv`](https://github.com/SysBioChalmers/Human-GEM/blob/main/model/metabolites.tsv)). For example, if some reactions and/or metabolites are removed from model, the corresponding lines should also be moved from the annotation tsv files and to `deprecatedReactions.tsv` and/or `deprecatedMetabolites.tsv` files under `/data/deprecatedIdentifiers` folder.

pranasag:
@Hao-Chalmers my bad, I'll fix that in the next hours, thanks for the explanation!

Hao-Chalmers:
@pranasag no worries.

pranasag:
Please note I also removed the gene `ENSG00000143891` which becomes orphan when reaction `MAR07745` is removed.

JonathanRob:
Thanks for your help, @pranasag.

I tend to agree with your logic, in that it doesn't really add anything meaningful to the model to have parallel reaction pathways for the different stereoisomers, since they don't produce anything different in the end (other than a few additional isomeric compounds). This is a bit similar to how we deal with lipid metabolism, where we don't try to model every possible combination of metabolites and reactions, but instead keep the most representative, common, or generic pathway form. Finally, having parallel reaction pathways can be problematic if simulations can seemingly compensate with the alternative pathway, when in reality, it may exhibit very poor efficiency. 

On the other hand, I share @Hao-Chalmers's concerns of removing content/information from the model unless really necessary.

In this case, I would recommend removing the reactions as you suggest, @pranasag, except for the reaction transforming between the two glucose isomers (glucose and beta-glucose). This way, we still retain the information that this transformation can take place (and thus don't lose a gene association), but don't retain unnecessary components which are effectively a duplication of existing metabolites/reactions. I realize this creates a dead-end reaction and metabolite, but our approach with Human-GEM has been to allow dead-ends since they often contain valuable information and can easily be removed if not needed (e.g., for simulation).

What do you think?

pranasag:
Good one, @JonathanRob, thanks for jumping in! 
If creating dead-ends in favor of retaining information is still acceptable, I see your suggestion as a win/win option out of this. So I'll revert the commits and only remove beta-D-G6P metabolite and related reactions, leaving the conversion of glucose in the model.

mihai-sysbio:
> Only few models from BIGG database have additional species (such as [this](http://bigg.ucsd.edu/universal/metabolites/glc__bD), [this](http://bigg.ucsd.edu/universal/metabolites/b_D_glucose) or [this](http://bigg.ucsd.edu/universal/metabolites/a_D_glucose)), while vast majority has only [D-glucose](http://bigg.ucsd.edu/universal/metabolites/glc__D).

@pranasag on a side note, I'm wondering is this might be an artifact of several models being based on the same template. Similarly, I see that the stereoisomers exist in Recon3D with a reference to HMR2.

Hao-Chalmers:
> I realize this creates a dead-end reaction and metabolite, but our approach with Human-GEM has been to allow dead-ends since they often contain valuable information and can easily be removed if not needed (e.g., for simulation).

@JonathanRob thanks for the clarification (this should be added to the Wiki as guidelines for future). Besides being a GEM, Human-GEM serves as a knowledgebase that allows dead-end reactions, which may be useful in other applications (e.g. integrative analysis).

According to a previous issue #44, unnecessary reactions should be initially deleted `softly` by blocking them `ub=lb=0`. While hard deletion will be implemented later in a major release (e.g. Human 2.0), this came from lessons learned from previous cases of first deleting and then adding back later. Should this soft deletion approach be applied here, what do you think? @pranasag

JonathanRob:
Just to comment, @Hao-Chalmers, the "soft deletion" approach was developed before we had implemented the yaml workflow, when it was much harder to track and revert changes to the model (in particular deletion of components). Now that we have the convenient yaml-based workflow, I wouldn't see a need to use the soft deletion approach.

Hao-Chalmers:
@pranasag I wonder if you also notice other `beta-` version metabolites than beta-D-glucose?

pranasag:
@Hao-Chalmers there's no extra metabolites for fructose and mannose, but for galactose, there's `galactose-1-phosphate[c]` (`MAM01911c`) and `alpha-D-galactose-1-phosphate[c]` (`MAM01322c`). Some of the reactions are duplicated for these two species, but `MAR08767` is unique to the `alpha` form, so, resolving this completely would require to remove reactions with `alpha-D-galactose-1-phosphate[c]` and to modify `MAR08767` (substitute with `MAM01911c`).

Hao-Chalmers:
> Some of the reactions are duplicated for these two species, but MAR08767 is unique to the alpha form, so, resolving this completely would require to remove reactions with alpha-D-galactose-1-phosphate[c] and to modify MAR08767 (substitute with MAM01911c).

The reaction [MAR08767](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR08767) is associated with gene [GALT](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000213930) (ENSG00000213930), which encode Galactose-1-phosphate uridylyltransferase (EC 2.7.7.12) that catalyzes the conversion of galactose-1-phosphate to glucose-1-phosphate in the Leloir pathway. According to [UniProt](https://www.uniprot.org/uniprot/P07902), [Rhea](https://www.rhea-db.org/rhea/13989), [MetaCyc](https://biocyc.org/META/NEW-IMAGE?type=REACTION&object=GALACTURIDYLYLTRANS-RXN) and [BRENDA](https://www.brenda-enzymes.org/enzyme.php?ecno=2.7.7.12#GENERAL%20INFORMATION), all the reactants and products are and only in alpha form.

In addition, there is on literature found by searching Pubmed with key words `beta-D-galactose-1-phosphate` and `GALT`. While two papers ([271990](https://www.pnas.org/content/74/12/5628) and [2011574](https://www.pnas.org/content/88/7/2633)) can be found by searching `alpha-D-galactose-1-phosphate` and `GALT`, both described the use of alpha form.

In sum, there is no evidence supporting the change of alpha form in MAR08767.

JonathanRob:
I'm not sure I fully understand what the consensus is at this point. Would this interpretation be correct, or did I misunderstand?
1) Remove glycolysis detour (except for MAR07745 which converts between alpha and beta glucose)
2) Leave the galactose pathway as-is

pranasag:
@JonathanRob even though I'd argue that the issue I see with the galactose pathway is very similar to the one for glycolysis (from the modelling perspective, while agreeing to the experimentally-relevant observations @Hao-Chalmers collected), maybe then let's stick to clearing up glycolysis at this time?

Hao-Chalmers:
> I'm not sure I fully understand what the consensus is at this point. Would this interpretation be correct, or did I misunderstand?
>
> 1. Remove glycolysis detour (except for MAR07745 which converts between alpha and beta glucose)
> 2. Leave the galactose pathway as-is

@JonathanRob makes sense to me

> let's stick to clearing up glycolysis at this time?

@pranasag agree

Hao-Chalmers:
Revised implementation plan after GPR checking:
- Remove `beta-` form metabolites MAM01388c and MAM01389c
- MAR07745: retain as-is
- MAR07746: remove, and modify GPR of its counterpart (MAR04394) by taking away ADPGK (which is ADP-dependent)
- MAR07747: retain, but change the form beta to generic
- MAR07748: remove, seems no evidence supporting this form conversion reaction
- MAR07749: remove, duplicate to MAR04381

@pranasag @JonathanRob what do you think?

pranasag:
Well done @Hao-Chalmers, sounds great! Also thanks for noticing `MAR07747` being ADP-dependent, I accidentally overlooked this.

JonathanRob:
Looks great, @Hao-Chalmers

Hao-Chalmers:
@pranasag  how about restarting a new PR to implement the revised plan, this probably is easier than making amendments to the current one?

Instead of pushing from a forked repo, I suggest making new PR directly to Human-GEM (you've assigned with write access), so that we can help in case there are issues raised by GitHub actions.

pranasag:
Thanks @Hao-Chalmers, I'll make a branch, do the changes and submit a new PR then.

mihai-sysbio:
@Hao-Chalmers what was the reason to keep `MAM01388c` in the model in the [revised plan](https://github.com/SysBioChalmers/Human-GEM/pull/307#issuecomment-911518585)?

JonathanRob:
@mihai-sysbio MAM01388c (beta-D-glucose) is part of the conversion reaction MAR07745, which we wanted to keep because it represents unique chemistry that is not duplicated elsewhere. Removing this would remove information (and a gene association) from the model.

Issue #310, opened 2021-09-01
MAM03165c sedoheptulose, wrong xref


ANiknejad:
Hello,

tricky issue

https://metabolicatlas.org/explore/Fruitfly-GEM/gem-browser/metabolite/MAM03165c

the xref to chebi is correct

https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:16802

but the xref to KEGG is to 
https://www.genome.jp/dbget-bin/www_bget?C02076

and this KEGG has xref to 2 chebi ids:
https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:16802 ---> OK
but also
https://www.ebi.ac.uk/chebi/searchId.do?chebiId=CHEBI:27520
which is the pyranose form, and makes things complicated

CHEBI has this information
β-D-sedoheptulopyranose (CHEBI:27520) is a sedoheptulose (CHEBI:16802) 

In KEGG, 2 compounds exist
https://www.genome.jp/entry/C08355

https://www.genome.jp/entry/C02076

but the xref that are present into
https://www.genome.jp/entry/C02076
are maybe wrong, because merging CHEBI:16802 and CHEBI:27520

In summary, I guess, it would be better for MAM03165c not to have xref to KEGG but only to CHEBI:16802

Thoughts?

Hao-Chalmers:
Agree

Hao-Chalmers:
Problem fixed in #312, good job @ANiknejad

PR #312, opened 2021-09-02
fix: annotation for MAM03165c

ANiknejad:
see issue #310


Hao-Chalmers:
@ANiknejad please modify this PR by changing the target branch from `SysBioChalmers:main` to `SysBioChalmers:develop`

JonathanRob:
The only change here is to the `metabolites.tsv` file, yet the "YAML validation" check is failing... why is that? @mihai-sysbio or @Hao-Chalmers.

Hao-Chalmers:
It's strange that the error message is `Gene annotation mismatch!`, but this PR has nothing to do with genes ...

mihai-sysbio:
Previously, I have encountered cases where aggressive caching led to workflow failures. As expected, running this check locally returns:
```
All checks have passed.
```
After having requested several re-runs of this workflow leading to failures, I would be okay with merging as-is.

Hao-Chalmers (approved):

PR #321, opened 2021-09-10
Human 1.10

Hao-Chalmers:
### Main improvements in this PR:
- Fix (PR #324):
	-  force cache refreshing by using commit SHA as key, to address random GH action failures
- Style (PR #320):
	- simplify `contributing guidelines` to adapt with yaml workflow
- Features (PR #315):
	- expand to cover all participants who have constructively contributed to `Human-GEM`, as claimed in #288
- Fixes (PR #313):
	- removed beta-D-glucose detour, based on the discussion in #307
		- MAM01389c, MAR07746, MAR07748, MAR07749 are removed
		- MAR07747 changed stoichiometry from `beta-` form to generic
- Fixes (PR #312):
	- remove KEGG annotation to MAM03165c, as proposed in #310
- Features (PR #305):
	- Include `Cellfie's` metabolic tasks, as discussed in #280
		- all metabolite ids are adjusted for `Human-GEM` except `Try_ggn`
		- change compartment `[x]` into `[s]`
		- metes that are both in `IN` and `OUT` columns (e.g. H20) are removed from `OUT`
- Fixes (PR #302):
	- removed ChEBI annotation `71980` to MAM00535c, as suggested in #273
- Features (PR #298):
	- adopt [`all-contributors`](https://allcontributors.org/) to recognize those who contributed and improve their visibility

PR #323, opened 2021-09-13
feat: enforce github actions checkout depth

mihai-sysbio:
This PR aims to try different ways of avoiding workflow failures as in https://github.com/SysBioChalmers/Human-GEM/pull/321.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -13,6 +13,8 @@ jobs:
 
     - name: Checkout
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 1
 
     - name: YAML Lint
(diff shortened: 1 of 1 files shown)


mihai-sysbio:
While this should not, in theory, make a difference, I'm wondering if this will resolve the caching issues that I suspect are the source of the failures in the Actions.

Hao-Chalmers:
@mihai-sysbio could the commit in this PR be cherry picked to #321 and see if it makes a difference?

mihai-sysbio:
@Hao-Chalmers triggering this workflow requires a PR, so I ended up mirroring this in #324 .

Hao-Chalmers:
The same error recurred in #324, this probably means that there is a flaw in the `develop`?

PR #324, opened 2021-09-13
fix random workflow failures

mihai-sysbio:
Local equivalent of #323, aiming to fix the apparently random unexpected `yaml-validation` workflow failures.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -14,6 +14,13 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
+    - name: Cache model folder
+      id: cache-model-files
+      uses: actions/cache@v2
+      with:
+        path: model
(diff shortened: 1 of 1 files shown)


mihai-sysbio:
Commit 1a8041c adds a caching mechanism where the model folder is cached with the commit SHA as the key, thus leading to cache refresh with every new commit. This seems to have resolved the problem.

Hao-Chalmers:
then if the increase to `fetch-depth` is still necessary?

mihai-sysbio:
@Hao-Chalmers a depth of `1` is the default value, so I edited the history of this branch to remove that commit.

Hao-Chalmers (approved):
Base on the comparison between #323 and #324, it appears that the recent GH action failure is cache-related issue and can be fixed by forced cache refreshing using commit SHA as key.

Issue #314, opened 2021-09-03
Why all compartment ids are not consistent?


smoretti:
Hi

I wonder why "c_i" is the only compartment of the model that is not a single letter?
**i** exists in BiGG (http://bigg.ucsd.edu/compartments/i) and could be used like this instead of c_i

Also in BiGG the extracellular compartment is *e*, not s that is eyespot.
Would you mind to change them?

mihai-sysbio:
Thanks for getting in touch @smoretti . Could I ask where did you encounter `c_i`? I'm only asking because I couldn't find it in the `yml` file. Here is the `compartments` section:
```
- compartments: !!omap
    - s: "Extracellular"
    - p: "Peroxisome"
    - m: "Mitochondria"
    - c: "Cytosol"
    - l: "Lysosome"
    - r: "Endoplasmic reticulum"
    - g: "Golgi apparatus"
    - n: "Nucleus"
    - i: "Inner mitochondria"
    ```

smoretti:
I found it in the SBML file

Hao-Chalmers:
> Also in BiGG the extracellular compartment is e, not s that is eyespot.
> Would you mind to change them?

this might be a good idea

Hao-Chalmers:
> I wonder why "c_i" is the only compartment of the model that is not a single letter?

This probably was due to an issue of the SBML exporting function, but should be fixed in the next release.

mihai-sysbio:
The `i` to `c_i` conversion is reported as a [bug in RAVEN](https://github.com/SysBioChalmers/RAVEN/issues/366).

Hao-Chalmers:
the `c_i` issue has been fixed in the latest release [v1.10](https://github.com/SysBioChalmers/Human-GEM/releases/tag/v1.10.0)

Hao-Chalmers:
> Also in BiGG the extracellular compartment is `e`, not `s` that is eyespot. Would you mind to change them?

In addition to the BIGG models, [yeastGEM](https://github.com/SysBioChalmers/yeast-GEM/blob/main/model/boundaryMets.txt) also use `e` as abbreviation of extracellular compartment. Therefore it makes sense to apply it to Human-GEM.

The modification to model file would be simple and straightforward. But I'm afraid this would affect some functions in HumanGEM and RAVEN. @JonathanRob what do you?

JonathanRob:
[...]

PR #330, opened 2021-10-15
feat: change compartment abbreviation

smoretti:
**Main improvements in this PR:**

- change Extracellular compartment abbreviation from `s` to `e`, as discussed in #314. Extracellular compartment and chemical species ids have been updated.

- [x]  Selected `develop` as a target branch


mihai-sysbio:
@Hao-Chalmers could you share more details on what more work is required? I read that @JonathanRob mentioned backwards compatibility, but it's hard for me to know specifically where that needs to be addressed.

JonathanRob:
@mihai-sysbio one of my concerns with changing the compartment abbreviation was actually whether this would affect Metabolic Atlas in any way? Since we need to change the actual metabolite IDs, it's not quite as straightforward as if we were changing an external/associated ID.

We also need to determine what to do with the existing IDs. Should they be added to the `metRetired` column in the `metabolites.tsv` file? Or just change them and not modify anything else in the `metabolites.tsv` file (except the metabolite ID, of course)?

We would then just need to check what other functions or data files have hard-coded the compartment abbreviations somewhere within. I don't think there are many cases, but the metabolic task files would of course be one case.

JonathanRob:
@mihai-sysbio ok great suggestions.

> Therefore, my suggestion is to deprecate these identifiers as if they would not be replaced, i.e. to copy the corresponding lines to deprecatedMetabolites.tsv and then do the renaming.

Good idea, we can plan to do that.

> There didn't seem to be anything interesting in the `code` folder pertaining to `s`

My guess is that `x` would be more of an issue. For example, the [`addBoundaryMets`](https://github.com/SysBioChalmers/Human-GEM/blob/main/code/addBoundaryMets.m) function should be updated.

> Should this PR also include the renaming of x to something else (b?) and p to x ?

Yeah, definitely. We should do all the renaming of these at once. I think a change of boundary to `b` would be fine, unless there's some other compartment where `b` would be the more intuitive letter.

Hao-Chalmers:
> @Hao-Chalmers could you share more details on what more work is required? I read that @JonathanRob mentioned backwards compatibility, but it's hard for me to know specifically where that needs to be addressed.

@mihai-sysbio thanks for bring this up. Specially, things should be addressed are, but not limited to:
- align metabolic task files to the change from `s` to `e`
- fix function `addBoundaryMets` while changing `x` to `p`

Hao-Chalmers:
Very good discussions @JonathanRob @mihai-sysbio 

Agree with the proposed plan (copy the corresponding lines to `deprecatedMetabolites.tsv` and then do the renaming). Let's focus on the change from `s` to `e` in this branch and including modification of metabolic task files as well

JonathanRob:
> Let's focus on the change from s to e in this branch 

@Hao-Chalmers I would recommend that we make all the compartment changes together in one branch, or is there some reason that you would prefer to break it up?

JonathanRob:
If it is OK with all of you @smoretti, @Hao-Chalmers, @mihai-sysbio, we can close this PR and I will implement abbreviation changes of all the compartments (extracellular, peroxisome, and boundary) in a new, single branch. This would also avoid having to update this branch with `develop`, which it's currently pretty far behind.

Hao-Chalmers:
> If it is OK with all of you @smoretti, @Hao-Chalmers, @mihai-sysbio, we can close this PR and I will implement abbreviation changes of all the compartments (extracellular, peroxisome, and boundary) in a new, single branch. This would also avoid having to update this branch with develop, which it's currently pretty far behind.

@JonathanRob  sounds good, go ahead please.

Hao-Chalmers:
> @Hao-Chalmers I would recommend that we make all the compartment changes together in one branch, or is there some reason that you would prefer to break it up?

based on lessons learned from previous work, here are some advantages to have each branch with more specific changes:
- clear goal and simple task
- quicker implementing and easier reviewing
- lower cost if made mistakes
- faster looping back from issue/discussion to the merging of implementation PR

mihai-sysbio:
Even though all changes are based on the same rationale, and maybe some of the manual checks will be done only once, it was a good idea to split the work. That being said, if it makes @JonathanRob's life easier, it's fine by me.

JonathanRob:
Ah ok, I think both of you @Hao-Chalmers and @mihai-sysbio make good points about splitting it up, but for simplicity I will combine them into a single branch/PR.

JonathanRob:
Closing this and implementing the changes in #341

Issue #311, opened 2021-09-01
In essential metabolic tasks: NEFA blood pool in[x] should be changed for NEFA blood pool in[s]


CadavidJoseL:
Essential task 46 (Uptake and beta oxidation of all NEFAs) fails in version 1.9. NEFA blood pool in[x] should be changed for NEFA blood pool in[s]; this solves the issue.

JonathanRob:
@CadavidJoseL Thanks for finding this, it should be changed to the [s] compartment as you suggest.

However, this task should not fail, even with the [x] compartment being used. Did you remember to add the boundary metabolites before running `checkTasks`?:

```
model_boundary = addBoundaryMets(model);
```

If the boundary metabolites are not present, then the task check results will not be valid.

Hao-Chalmers:
@CadavidJoseL Thanks, will look into this

CadavidJoseL:
You're right, I did forget to add those boundary mets; that also fixes the issue. Thanks!

JonathanRob:
No problem, but I believe you are still correct in that the `NEFA blood pool in` compartment should be changed from [x] to [s] in the task file.

JonathanRob:
If it's ok with you, @CadavidJoseL, I'd like to keep this issue open until we correct the compartment of the `NEFA blood pool in` in the task file. Otherwise we may forget to make the fix.

JonathanRob:
This issue was resolved in #331.

PR #331, opened 2021-11-05
feat: taskfile updates

JonathanRob:
Converts all metabolic task files to `.txt` format to allow easier git-based tracking.

- All metabolic task files in `/data/metabolicTasks` are converted from `.xlsx` to `.txt`.
- Functionality was confirmed by loading each of the file formats by loading into Matlab with `parseTaskList` and confirming that the two loaded structures are identical (`isequal` yields `True`).
- The original `.xlsx` versions of the task files are removed to avoid risk of the two file formats becoming out of sync when changes are made to any of the task files.
- This PR also addresses #311 by correcting the compartment of `NEFA blood pool in` from Boundary to Extracellular (`[x]` to `[s]`)
- A minor fix is applied to `translateGrRules` which adds some flexibility to input parameter recognition.


mihai-sysbio:
At a quick glance some tasks exist both in the `Essential` and `Full` lists. For the same reason that duplication increases chances of errors (thus `txt` only, no more `xlsx`), I'm wondering if `Full` couldn't be then obtained by merging the existing `Essential` with new `NonEssential` list.

Hao-Chalmers:
it seems Matlab license expires in GH actions @mihai-sysbio

Hao-Chalmers:
it's a good move to replace `xlsx` task files with `txt` ones.

JonathanRob:
> it seems Matlab license expires in GH actions @mihai-sysbio

Ah ok, so that's why I got like 6 emails from GitHub about failed checks...

JonathanRob:
> At a quick glance some tasks exist both in the `Essential` and `Full` lists. For the same reason that duplication increases chances of errors (thus `txt` only, no more `xlsx`), I'm wondering if `Full` couldn't be then obtained by merging the existing `Essential` with new `NonEssential` list.

Good point, this is something we've noticed earlier. Merging would probably be a good idea, though it will require some manual work since some of the tasks that have the same/similar name are not implemented in exactly the same way.

mihai-sysbio:
To resolve the remaining Action failures, I propose merging https://github.com/SysBioChalmers/Human-GEM/pull/332 into `develop`, and then updating this branch.

Hao-Chalmers:
the Action failures have been fixed after merging #332. Good job @mihai-sysbio

mihai-sysbio:
@Rasools this PR is ready for review 😊

code/GPRs/translateGrRules.m
```diff
@@ -117,6 +117,9 @@
 
 % allow for non-exact matches to "Symbol" option
 targetFormat(ismember(lower(targetFormat),{'name','names','symbol','symbols'})) = {'Symbol'};
+if ismember(lower(gene_type_orig), {'name','names','symbol','symbols'})
```
Rasools, line 120: Hi @JonathanRob, I saw you have shortly explained this change in the PR #331 conversation tab, I was wondering if this change is connected to any issue that could explain the idea behind it? thanks

JonathanRob: Good question @Rasools. It's not connected to any particular issue. It's just based on my own experience using the function, since I never remember whether it's "name" or "symbol" or pluralized, etc. so I assumed others might run into a similar issue.

Rasools: Great! thanks for adding this.

Rasools (approved):

Hao-Chalmers:
it's strange that this PR cannot be merged after the approval of @Rasools, why is this?

I've gone through all the changes, and all TXT tasks file have identical content to their counterparts in Excel (before commit 2ff239b for the essential task file) after parsing.

Hao-Chalmers (approved):

PR #334, opened 2021-11-08
refactor: remove recon ids from reaction names

mihai-sysbio:
This PR removes the Recon ids from the reaction `name` field. As far as I can tell, these are anyway kept in `reactions.tsv`.


Hao-Chalmers (approved):

mihai-sysbio:
In VS Code I used Find and Replace with this regex
```
(- id: "MAR\d+"\n +- name: ")RE\d\d\d\d.
```
to replace with
```
$1
```

The way this works is to first finding all reaction (starting with MAR) where the name matches the `RE..` pattern. Then, the first part is surrounded by `( )` to become a "captured" group, later to be referenced by `$1`. This basically replaces the entire find pattern with the first part of the find pattern, thus effectively deleting what is outside the `( )`: only the `RE..` name.

PR #341, opened 2021-12-19
feat: change multiple compartment abbreviations

JonathanRob:
### Main improvements in this PR:
This is a more extensive/exhaustive implementation of #330 to address discussions in #314.

Three Human-GEM compartment abbreviations are changed in the model file, metabolite annotation file, metabolic task files, and the `addBoundaryMets` function:

| Compartment | Abbreviation | New Abbreviation
| --- | --- | --- |
| Extracellular | `s` | `e` |
| Boundary | `x` | `b` |
| Peroxisome | `p` | `x` |

Since the changes (diffs) are quite large, I implemented the changes with a script `changeCompAbbrevs` in `code/modelCuration/`.

**I hereby confirm that I have:**

- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch

PR #343, opened 2022-01-14
Update comps in metabolites.tsv to resolve branch conflict

JonathanRob:
Non-elegant solution to resolve merge conflicts with #341 in the `metabolites.tsv` file. This branch updates the metabolite abbreviations to be consistent with the new abbreviations (x -> b, s -> e, p -> x).


JonathanRob:
The checks will fail on this PR because the model file is now out of sync with the metabolites annotation file. I think it should anyway be OK to merge.

mihai-sysbio (approved):
Looks good to me, but it's really hard to verify in detail. On the surface (number of changes; line count), things seem to make sense. I will do a more thorough review on #341.

PR #361, opened 2022-02-14
fix: broken README

### LINKED ISSUE #360: fix: all contributors formatting in Readme
### Description of the issue:
The README appears broken for some time.

A README with emojis.

mihai-sysbio:
This PR resolves #360.
The core issue stems from the different README files that exist on branches `main` vs `develop`. The reason for having different files is that the README contains stats on model component counts. 

This fix is targeting the `main` branch, resolving only the differences in:
- the spelling of Pinar's family name
- the `all-contributors` section, which was the problem identified in #360

My suggestion for a different PR would be to replace the stats section with a link to Met Atlas that includes these stats, thus not needing to have different README files in the different branches, and in consequence to also remove the function that does this replacement. Such a PR would therefore need to target `develop`, unlike this one.


haowang-bioinfo (approved):

mihai-sysbio:
> My suggestion for a different PR would be to replace the stats section with a link to Met Atlas that includes these stats, thus not needing to have different README files in the different branches, and in consequence to also remove the function that does this replacement. Such a PR would therefore need to target `develop`, unlike this one.

@haowang-bioinfo any thoughts on this?

PR #362, opened 2022-02-17
fix: remove gpr for MAR04840 resolves #342

### LINKED ISSUE #342: I suspect that the GPR for MAR04840 is wrong and should be removed
### Description of the issue:
MAR04840 {'H2O[c] + H2S2O3[c] => 2 H+[c] + H2S[c] + sulfate[c]'} currently has the GPR ENSG00000198610 (AKR1C4). Comment here https://www.genome.jp/entry/R00864 says it is a spontaneous reaction. And, when I look at the gene (https://www.proteinatlas.org/ENSG00000198610-AKR1C4) and EC number (https://www.brenda-enzymes.org/enzyme.php?ecno=1.1.1.50), these describe a totally different reaction, involving NAD+/NADP+? 

I suspect that the EC number and GPR should be cleared, and that the reaction should be marked as spontaneous?

- [ ] Done this analysis in the `main` branch of the repository
- [ ] Checked that a similar issue does not exist already

mihai-sysbio:
This PR resolves #342 by applying the suggestion to remove the gene (and thus the GPR) for `MAR04840`.


mihai-sysbio:
> As I understand it, the reaction has not been made spontaneous in reactions.tsv? This matters for some things, such as the new tINIT and Gecko Light.

I've checked the MNX identifier, and it was marked as spontaneous there. The `reactions.tsv` contains only annotation, so I'm not sure how that file would play a role.

johan-gson (approved):

Issue #354, opened 2022-01-27
RNA formation


PkiwiBird:
### Description of the issue:
1. reaction MAR07161, representing gene transcription (RNA formation) is not happening in the nucleus
2. there is no RNA in the nucleus and transport reaction of RNA from the nucleus to the cytoplasm
3. not sure if reaction MAR07162 should exist as RNA formation from ribonucleoside diphosphates (instead of triphosphates) seems to have mainly happened for primordial nucleic acid biosynthesis ([study here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6682740/)). 
Also, it is annotated to enzyme [2.7.7.8 in the metabolic atlas](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR07162), which seems to catalyze the reverse reaction: [brenda](https://www.brenda-enzymes.org/enzyme.php?ecno=2.7.7.8), [keeg](https://www.genome.jp/entry/2.7.7.8). In humans that enzyme seems to be involved in mRNA degradation in the cytoplasm: "In the cytoplasm, shows a 3'-to-5' exoribonuclease mediating mRNA degradation activity"(from [here](https://www.uniprot.org/uniprot/Q8TCS8))

#### Expected feature/value/output:
1. reaction MAR07161 could be moved to the nucleus, as gene transcription occurs in nucleus in humans.
2. model could have an RNA[n] metabolite and create a transport reaction from the nucleus to the cytoplasm with bounds 0 to 1000 as RNA does not re-enter the nucleus once transported to the cytoplasm.
3. maybe remove reaction MAR07162 or maybe allow it to be reversible?

#### Python code:
```import os
import cobra
from cobra import Metabolite, Reaction
from urllib.request import urlretrieve

# load human1.10 and save:
if not os.path.exists('data/models/human1_10.xml'):
    mdPthSBML='https://raw.githubusercontent.com/SysBioChalmers/Human-GEM/master/model/Human-GEM.xml' # path to generic SBML model
    modelPathSBML, _ = urlretrieve(mdPthSBML)
    m = cobra.io.read_sbml_model(modelPathSBML)
    cobra.io.write_sbml_model(m, 'data/models/human1_10.xml')
else:
    m = cobra.io.read_sbml_model('data/models/human1_10.xml')

# fva to identify blocked reactions:
blck_rc = cobra.flux_analysis.find_blocked_reactions(m)
len(blck_rc) # 1190

[...]

PR #364, opened 2022-02-17
refactor: RNA formation, issue #354

PkiwiBird:
implement the suggestions of issue #354 
- add compartmentalised metabolite RNA[n] (MAM02847n)
- change reaction MAR07161 (RNA creation) from cytoplasm to nucleus
- add transport reaction of RNA from the nucleus to the cytoplasm (MAR13086)
- move gene rule of reaction MAR08639 to reaction MAR13086, as it represents the nuclear pore complex
- swap the direction of reaction MAR07162, as explained in issue #354
- remove the reaction of DNA transport between the nucleus and cytoplasm MAR08639, since DNA shouldn't go out of the nucleus and the reaction's gene rule was already associated to another reaction, MAR13086

Comments:
- not sure what is the appropriate reaction score for reaction MAR13086
- and could not find ids form other databases for this reaction, probably beacause RNA has to be first processed into mature RNA (with polyA-tail and 5'Cap) before moving to cytoplasm...


model/reactions.tsv
```diff
@@ -7191,7 +7191,6 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "MAR08412"	""	"O2Stn"	""	""	""	"O2Stn"	"MNXR102089"	"HMR_8412"	"RCR20749"	""	0	""	""	"HMR_8412"
 "MAR08447"	""	"CYTK1"	""	""	""	"CYTDtn"	"MNXR97047"	"HMR_8447"	"RCR20750"	""	0	""	""	"HMR_8447"
 "MAR08476"	""	""	""	""	""	"HMR_8476"	""	"HMR_8476"	"RCR21086"	"2.A.57.1.4"	0	""	""	"HMR_8476"
-"MAR08639"	""	""	""	""	""	"HMR_8639"	""	"HMR_8639"	"RCR21087"	""	0	""	""	"HMR_8639"
```
JonathanRob, line 7194: Could you please add this deleted reaction (`MAR08639`) to the `deprecatedReactions.tsv` file (in `data/deprecatedIdentifiers/`)? You can just copy the deleted row from `reactions.tsv` and append it to the end of `deprecatedReactions.tsv`.

JonathanRob (changes requested):
> not sure what is the appropriate reaction score for reaction MAR13086

Eh, good question. I think you can use the same score as was used for MAR08639, which was a 2.

> and could not find ids form other databases for this reaction, probably beacause RNA has to be first processed into mature RNA (with polyA-tail and 5'Cap) before moving to cytoplasm...

Yeah, it's a very simplified representation of the process, so if you were not able to find anything from other databases after searching for a bit, then I think it's OK to leave it as is.

model/reactions.tsv
PkiwiBird: I think it is ok now

JonathanRob (approved):

PR #388, opened 2022-05-13
ftINIT helpers

johan-gson:
*Added help functions that are used together with ftINIT in RAVEN. A pull request containing the ftINIT implementation has been done separately in RAVEN, but the changes should not be dependent on each other.


code/tINIT/getAATripletReactions.m
```diff
@@ -0,0 +1,69 @@
...
+%
+%   model               The model to be examined.
+%   onlyRxnsWithoutGPRs If true, the function only gets the rxns without GPRs
+%
+%   rxns                The resulting reactions
+%
+%   Usage: rxns=getAATripletReactions(model, onlyRxnsWithoutGPRs)
+%
+
+aminoAcidsCap = {'Alanine';'Arginine';'Asparagine';'Aspartate';'Cysteine';'Glutamine';'Glutamate';'Glycine';'Histidine';'Isoleucine';'Leucine';'Lysine';'Methionine';'Metheonine';'Phenylalanine';'Proline';'Serine';'Threonine';'Tryptophan';'Tyrosine';'Valine'};
```
JonathanRob, line 14: Not sure what `Metheonine` is, but it can probably be removed (should only be 20 AAs)

johan-gson: So, the problem is that there are reactions where the amino acids are misspelled - that's why I have added those weird ones :)

mihai-sysbio: How about submitting another PR that fixes the misspelling, and merging that one first?

JonathanRob: Yeah sorry, I should have actually looked into it further before commenting.

mihai-sysbio: Is there a reason to not code in [all 22](https://en.wikipedia.org/wiki/Essential_amino_acid#Essentiality_in_humans)?

johan-gson: So, the point of this function is to identify reactions that convert amino acid triplets back and forth, i.e. chains of three amino acids <=> three separate amino acids. It is a measure of optimization - these reactions are pretty pointless in most practical uses with tINIT. It is like 700 reactions or something like that. So I only added the amino acids that help identify those reactions.

code/tINIT/getINITModel2.m
```diff
@@ -1,5 +1,7 @@
 function [model, metProduction, essentialRxnsForTasks, addedRxnsForTasks, deletedDeadEndRxns, deletedRxnsInINIT, taskReport] = getINITModel2(refModel, tissue, celltype, hpaData, arrayData, metabolomicsData, removeGenes, taskFile, useScoresForTasks, printReport, taskStructure, params, paramsFT)
 % getINITModel2
+%   This function is deprecated - we suggest that you instead use ftINIT in
+%   RAVEN Toolbox together with the helpers in this repo.
```
JonathanRob, line 4: I have mixed feelings about this. This function is still (nearly) the original implementation of the algorithm, and is not necessarily "deprecated". Although ftINIT is faster, it uses a different approach (with additional assumptions and simplifications). So unless we can definitively show that ftINIT is a more accurate algorithm, I hesitate to "deprecate" the original tINIT.

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -0,0 +1,126 @@
...
+    'MAR10033'; ...
+    'MAR10035'; ...
+    'MAR10036'; ...
+    'MAR10037'; ...
+    'MAR10038'; ...
+    'MAR10062'; ...
+    'MAR10063'; ...
+    'MAR10064'; ...
+    'MAR10065'; ...
+    'MAR13082'};
```
JonathanRob, line 98: There is a lot of hard-coded items in this function, which I generally would like to avoid. But since I don't really have better suggestions, I think it's fine to leave like this.

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -0,0 +1,126 @@
...
+    'MAR05277'; ...
+    'MAR05279'; ...
+    'MAR05281'; ...
+    'MAR05283'; ...
+    'MAR05291'; ...
+    'MAR09817'; ...
+    'MAR09818'};
+
+%reactions that just pool metabolites
+poolRxns = { ...
```
mihai-sysbio, line 74: Pool reactions are reactions that have `pool` in the name, so these could be defined dynamically.

code/tINIT/removeDrugReactions.m
```diff
@@ -0,0 +1,22 @@
+function outModel=removeDrugReactions(inModel)
+% outModel=removeDrugReactions(inModel)
+%
+% Removes reactions from Human-GEM related to drugs
```
JonathanRob, line 4: I wonder if it would be easier to just remove all reactions that are in the `Drug metabolism` subsystem. I imagine it might remove a few additional reactions, but unlikely that they would be important for anything. Something to think about in the future.

johan-gson: Aha, I never thought about that.

code/tINIT/getINITModel2.m
johan-gson: The question is what to write here. Should we say that there is a newer version, but not call it deprecated? I have some reason to believe that ftINIT is performing better. I'm supplying a powerpoint that shows one test I did with this function and compared it to tINIT without allowing flux in both direction and without metabolite secretion. ftINIT in this test gave the same result as the one where these were not allowed, while the standard getINITModel2 produced something that is quite different, full of gaps. The task step of ftINIT typically fills in ~13 rxns, while the same step for getINITModel2 fills in ~55 rxns. So, I think that getINITModel2 has some problems, and that the problems with ftINIT are much smaller. But I have not tested it super much.
[tINITTestInfo.pptx](https://github.com/SysBioChalmers/Human-GEM/files/8741931/tINITTestInfo.pptx)

JonathanRob: Ok, very good points. I think it's fine to state that there is a newer version, but would avoid using the word "deprecated", since it sometimes implies that the function will eventually be removed (which is not the intent).

I agree that some of the behavior you've seen with ftINIT (other than shorter compute time) can be interpreted as a clear improvement in performance/accuracy.

johan-gson: Ok, I wrote that there is a new improved version of the function available in RAVEN, I have pushed a new commit regarding this.

JonathanRob:
Ok, I'm still going through some of the details on this PR, though so far it looks ok to me. But why are the yaml conversion tests failing?

johan-gson:
Yeah, I don't get that either, I cannot see that I have done anything with that. What I understood from the error messages is that it fails to allocate some resource for running them?

mihai-sysbio:
> Ok, I'm still going through some of the details on this PR, though so far it looks ok to me. But why are the yaml conversion tests failing?

Now resolved.

JonathanRob (approved):

code/tINIT/removeDrugReactions.m
haowang-bioinfo: As a review, I have following concerns to this function:
- there is some commented lousy code
- it doesn't appear to be generic enough for future use, thus might be okay to be hard coded inside `prepHumanModelForftINIT`

johan-gson: Yeah, I'm not sure that should be there - this is to show how I tested that the code works. Maybe we could clean that up.

PR #392, opened 2022-06-03
refactor: turn `importYaml` into a more generic function

haowang-bioinfo:
* subsystem can be properly loaded regardless of the other reaction feature slots
* no longer sort gene list when importing Yaml file


edkerk:
I would suggest to have `importYaml` (and `exportYaml`) incorporated directly in RAVEN (together with #393), especially considerint you're working on making it more generic. There are also plans to use yaml file format to store GECKO models, so the improvements being made here would be beneficial for numerous projects.

mihai-sysbio (approved):
lgtm

Issue #387, opened 2022-05-05
importYaml is slow


johan-gson:
### Description of the issue:
Since the Human-GEM.mat file is removed from the repo we are down to using importYaml, which takes several minutes, while load('Human-GEM.mat') takes 1 s. If this is the intended use, it would be beneficial to speed this up. The way the code is written now, it performs tons of small reads from the file, which is super slow. The code could probably be enhanced a lot by reading the file in one big chunk into memory and then parse it from memory.

#### Expected feature/value/output:
It would just be much faster - right now it is annoyingly slow.

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [ ] Done this analysis in the `main` branch of the repository
- [ ] Checked that a similar issue does not exist already

johan-gson:
Aha, so the .mat file will still be available in the main branch. It is still a bit complicated, the code cannot really rely on that the file is there then, such as a tutorial, or can it?

mihai-sysbio:
Perhaps you're looking for [this tutorial](https://sysbiochalmers.github.io/Human-GEM-guide/getting_started)?

johan-gson:
Most useful, thanks! But it is still very slow :)

johan-gson:
When I look closer at it, it may not be the file reading, but it could be tested easily. The file reading could easily be fixed using readlines(filename), and just loop over the lines instead of checking for eof. But I'm not sure it will help that much. It could also be for example the regexp.

edkerk:
With `profile on; model=importYaml('../../model/Human-GEM.yml'); profile viewer` it becomes clear it is slow not because it is _read_ one line at a time, but because it is _processed_ one line at a time. More precisely, each field is growing by one entry each time.

line 289 is run 166181 times, taking 61 seconds in total:
```matlab
    model.(fieldName) = [model.(fieldName); {value}];
```

Refactoring of the code could greatly speed this up, if the function would instead: (1) read the whole file at once; (2) split the data in the part describing reactions and metabolites and process separately; (3) use regexp to find e.g. first line of each reaction; (4) use regexp to extract all entries for a certain field type (e.g. `eccodes`) all at once.

I can imagine that the tricky part will be to handle cases where e.g. not all reactions have an entry for a particular field. But with some clever coding, I'm sure this can be managed :).

But maybe you see some other opportunities?

johan-gson:
[...]

PR #393, opened 2022-06-14
refactor: speed up writing model fields

edkerk:
- refactor:
  - `importYaml` speed improvement, contributing to #387

On my PC, `importYaml('../../model/Human-GEM.yml');` tooks 248 seconds before the change and 12 seconds after the change.

### DIFF
--- a/code/io/importYaml.m
+++ b/code/io/importYaml.m
@@ -25,6 +25,25 @@ if ~(exist(yamlFilename,'file')==2)
     error('Yaml file %s cannot be found', string(yamlFilename));
 end
 
+if verLessThan('matlab','9.9') %readlines introduced 2020b
+    fid=fopen(yamlFile);
+    line_raw=cell(1000000,1);
+    while ~feof(fid)
+        line_raw{i}=fgetl(fid);
+        i=i+1;
+    end
+    line_raw(i:end)=[];
+    line_raw=string(line_raw);
+else
+    line_raw=readlines(yamlFile');
+end
+
+line_key=regexprep(line_raw,'^ *-? ([^:]+)(:).*','$1');
+line_key=regexprep(line_key,'(.*!!omap)|(---)','');
+
+line_value = regexprep(line_raw, '[^":]+: "?(.+)"?$','$1');
+line_value = regexprep(line_value, '"','');
+
 % Define the required fields of humanGEM
 % There are a total of 37 fields in the model so far, the non-generic ones
 % are excluded here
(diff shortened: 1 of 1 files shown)


mihai-sysbio (approved):
🦄

edkerk:
I couldn't stand not fixing the last part. It's now much faster, takes about 12 sec. I don't see room for more improvement.

code/io/importYaml.m
```diff
@@ -25,6 +25,25 @@
     error('Yaml file %s cannot be found', string(yamlFilename));
 end
 
+if verLessThan('matlab','9.9') %readlines introduced 2020b
+    fid=fopen(yamlFile);
```
johan-gson, line 29: A bit confusing with yamlFile and yamlFilename - it looks like yamlFilename is the right one, if I look at line 1?

code/io/importYaml.m
```diff
@@ -25,6 +25,25 @@
...
+    line_raw=cell(1000000,1);
+    while ~feof(fid)
+        line_raw{i}=fgetl(fid);
+        i=i+1;
+    end
+    line_raw(i:end)=[];
+    line_raw=string(line_raw);
+else
+    line_raw=readlines(yamlFile');
```
johan-gson, line 38: should this really be yamlFile' ?

johan-gson (changes requested):
Just take a look at the yamlFile things. I take it that you tested that it became faster?

code/io/importYaml.m
edkerk, line 29:
```suggestion
    fid=fopen(yamlFilename);
```

code/io/importYaml.m
edkerk, line 38:
```suggestion
    line_raw=readlines(yamlFilename');
```

edkerk: Fixed, thanks for noticing!

johan-gson (approved):
Now it looks better!

edkerk:
@haowang-bioinfo In f2d5352 I incorporated _similar_ changes as you introduced in #392 (but not exactly the same as it wouldn't match the refactoring that I did). Can you please check that it matches the behaviour that you need?

haowang-bioinfo:
> @haowang-bioinfo In https://github.com/SysBioChalmers/Human-GEM/commit/f2d5352f108986ef83a0262798e35602b03db20a I incorporated similar changes as you introduced in https://github.com/SysBioChalmers/Human-GEM/pull/392 (but not exactly the same as it wouldn't match the refactoring that I did). Can you please check that it matches the behaviour that you need?

will look into this

mihai-sysbio:
The `importYaml` function is used in the tests implemented with GH Actions. From the history I see the speedup is from [~5min30s](https://github.com/SysBioChalmers/Human-GEM/runs/6752909441?check_suite_focus=true) down to [~1m30s](https://github.com/SysBioChalmers/Human-GEM/runs/6903361764?check_suite_focus=true). With this confirmed speed improvement and tests passing, I will go ahead and merge this PR.

PR #411, opened 2022-08-24
fix: mark MAR08430 as spontaneous

### LINKED ISSUE #410: MAR08430 should be marked spontaneous
### Description of the issue:
MAR08430: creatine-phosphate[m] => creatine-phosphate[c]
According to wikipedia (https://en.wikipedia.org/wiki/Creatine_phosphate_shuttle, not a great source, but probably ok) this transport goes via diffusion, which means it is spontaneous. I think it is good to mark all spontaneous reactions since we don't need to look for missing GPRs then!

- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

mihai-sysbio:
This PR resolves #410.

### DIFF
--- a/model/reactions.tsv
+++ b/model/reactions.tsv
@@ -9868,7 +9868,7 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "MAR09924"	""	"EX_stcrn_e"	""	""	""	"EX_stcrn[e]"	""	""	""	""	0	""	""	"EX_stcrn[e]"
 "MAR09925"	""	"EX_pmtcrn_e"	""	""	""	"EX_pmtcrn[e]"	""	""	""	""	0	""	""	"EX_pmtcrn[e]"
 "MAR09926"	""	"EX_hdcecrn_e"	""	""	""	"EX_hdcecrn[e]"	""	""	""	""	0	""	""	"EX_hdcecrn[e]"
-"MAR09927"	""	"PCREATte"	""	""	""	"PCREATte"	""	""	""	""	1	""	""	"PCREATte"
+"MAR09927"	""	"PCREATte"	""	""	""	"PCREATte"	""	""	""	""	0	""	""	"PCREATte"
 "MAR09928"	""	"HC00342te"	""	""	""	"HC00342te"	""	""	""	""	0	""	""	"HC00342te"
 "MAR09929"	""	""	""	""	""	"C08261te"	""	""	""	""	0	""	""	"C08261te"
 "MAR09930"	""	"BGLYte"	""	""	""	"BGLYte"	""	""	""	""	0	""	""	"BGLYte"
--- a/model/reactions.tsv
+++ b/model/reactions.tsv
@@ -6718,7 +6718,7 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "MAR08071"	""	"SLDt"	""	""	""	"SLDt"	"MNXR104441"	"HMR_8071"	"RCR40927"	""	0	""	""	"HMR_8071"
 "MAR04312"	""	""	"RT0321"	""	""	"SBT_Dtde;SBTle"	""	"HMR_4312"	"RCR41640"	""	0	""	""	"HMR_4312"
 "MAR08428"	""	"CRMPte"	""	""	""	"CREATtmdiffir"	"MNXR96892"	"HMR_8428"	"RCR20534"	""	0	""	""	"HMR_8428"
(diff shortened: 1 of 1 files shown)


model/reactions.tsv
```diff
@@ -9868,7 +9868,7 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "MAR09924"	""	"EX_stcrn_e"	""	""	""	"EX_stcrn[e]"	""	""	""	""	0	""	""	"EX_stcrn[e]"
 "MAR09925"	""	"EX_pmtcrn_e"	""	""	""	"EX_pmtcrn[e]"	""	""	""	""	0	""	""	"EX_pmtcrn[e]"
 "MAR09926"	""	"EX_hdcecrn_e"	""	""	""	"EX_hdcecrn[e]"	""	""	""	""	0	""	""	"EX_hdcecrn[e]"
-"MAR09927"	""	"PCREATte"	""	""	""	"PCREATte"	""	""	""	""	0	""	""	"PCREATte"
+"MAR09927"	""	"PCREATte"	""	""	""	"PCREATte"	""	""	""	""	1	""	""	"PCREATte"
```
johan-gson, line 9871: Are you sure that creatine phosphate can diffuse out of the cell? I have not seen any evidence of this. It would probably not be desirable for muscle cells to lose it, I don't know how this works. My comment was for the mitochondrial membrane only, that is well known and part of the creatine phosphate shuttle.

haowang-bioinfo (approved):
looks good

model/reactions.tsv
JonathanRob: Ok, then maybe to be on the safe side, we only mark the mitochondria -> cytosol reaction as spontaneous for now?

johan-gson: That would be my recommendation

JonathanRob (approved):
I reverted MAR09927 (transport of creatine-phosphate between cytosol and extracellular) back to non-spontaneous since, as suggested by @johan-gson, we do not yet have enough evidence to mark it as spontaneous.

haowang-bioinfo:
looks good to me

mihai-sysbio:
Thanks @JonathanRob. In addition to doing that, I was also intending to add the reference used in the Wikipedia page to MAR08430.

mihai-sysbio:
Sorry for eagerly marking MAR09927 without looking at the compartments.

For the intended reaction MAR08430 I am having a hard time finding support for the diffusion in the cited article [PMID:3529851](https://pubmed.ncbi.nlm.nih.gov/3529851/). My suggestion would be, prior to merging, for this to be checked.

johan-gson:
Sorry for the mess, I wonder if I have not misunderstood this althogether. I think what they meant by the diffusion is that it diffuses to the fibrils from OUTSIDE mitochondria, so this is not related to actually passing through the membrane. Reading more on https://en.wikipedia.org/wiki/Creatine_kinase supports this - it seems that there are something called  PCr/Cr-shuttle that facilitate this. There should probably be a GPR here, but I didn't find anything when googling, so maybe we should just stop this and not assign anything as spontanous. If anyone has an idea of how to figure out the GPR, it would be nice though. Again, sorry for the mess.

haowang-bioinfo:
> it seems that there are something called PCr/Cr-shuttle that facilitate this. There should probably be a GPR here, but I didn't find anything when googling, so maybe we should just stop this and not assign anything as spontaneous.

No solid evidence found by browsing the citations listed in the wiki page of `Creatine phosphate shuttle`. Such shuttles usually should get involve with transporters, i.e. with GPR.

johan-gson:
Nice work Hao. So, maybe the reaction should be changed, is that what you are proposing? I'm still a bit confused by this. It seems there are many isozymes, some which operate in the cytosol (then generating ATP) and some that operate "in" mitochondria that consume ATP. Right now, I feel like I don't dare to change the reaction since things are still a bit unclear.

haowang-bioinfo:
To implement the schematic view, technically the reactions [MAR08427](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR08427) and [MAR08430](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR08430) should be modified by changing metabolites involved in `[m]` to `[i]`, while the GPR of MAR08427 probably should also be adjusted to the corresponding isozymes. These changes shouldn't affect model's performance, and would fit the commit did in this PR.

haowang-bioinfo:
@johan-gson look forward to your test results

johan-gson:
This will unfortunately have to wait, I only have 8 days left at SysBio and a lot of things to do :)

haowang-bioinfo:
Sure, all the best!

johan-gson:
@haowang-bioinfo I wonder if we shouldn't just cancel this, both the PR and the issue? My original idea was just not right, and I will not find any energy to dig into this anytime soon.

JonathanRob:
@johan-gson @haowang-bioinfo we can close for now and always re-open later if necessary.

haowang-bioinfo:
sure, let's close it

Issue #417, opened 2022-09-29
GPR curation for MAR04926


haowang-bioinfo:
### Description of the issue:
There are three genes associated with mitochondrial pyruvate transport reaction ([MAR04926](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR04926)):
1. [ENSG00000060762](https://www.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000060762) ([MPC1](https://www.uniprot.org/uniprotkb/Q9Y5U8/entry))
2. [ENSG00000143158](https://www.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000143158) ([MPC2](https://www.uniprot.org/uniprotkb/[O95563](https://metabolicatlas.org/identifier/UniProt/O95563)/entry))
3. [ENSG00000238205](https://www.ensembl.org/Homo_sapiens/Gene/Summary?db=core;g=ENSG00000238205) ([MPC1L](https://www.uniprot.org/uniprotkb/P0DKB6/entry))

According to this experimental [study](https://pubmed.ncbi.nlm.nih.gov/22628558/), mitochondrial pyruvate carrier is a complex, formed by two essential subunits: `MPC1` and `MPC2`, in the inner mitochondrial membrane. While there is no literature evidence associated with `MPCL1`, which shows 64.2% sequence similarity to `MPC1`.

#### Current GPR rule:
* `ENSG00000060762 or ENSG00000143158 or ENSG00000238205`

#### Expected GPR rule:
* `(ENSG00000060762 and ENSG00000143158) or (ENSG00000238205 and ENSG00000143158)`

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [x] Done this analysis in the `main` branch of the repository
- [x] Checked that a similar issue does not exist already

mihai-sysbio:
To me, ~65% sequence similarity is not enough to justify equivalent function, especially in the light of there being a meaningful difference in sequence length. The predicted structured in Alphafold do not seem to support the similarity. Thus, my suggestion would be, in absence of further evidence, to remove the association with `MPC1L`.

haowang-bioinfo:
> The predicted structured in Alphafold do not seem to support the similarity.

Excellent - can you manage to present the evidence of the predicted structure comparison between `MPC1` and `MPC1L`.

mihai-sysbio:
[...]

PR #419, opened 2022-09-30
fix: curate GPR rule for MAR04926

haowang-bioinfo:
* update GPR rule and reference PMIDs for `MAR04926`, as discussed in #417


mihai-sysbio (approved):
👍🏻 

To me, both old PMIDs looked to generic to be useful, but I haven't looked into detail if they should be removed or not.

PR #420, opened 2022-10-07
fix: remove some empty fields from imported yaml model

JonathanRob:
The `importYaml` function defines and creates a set of model fields at the start of import, which means that if some fields are missing from the `.yml` file, the corresponding fields in the Matlab structure will be empty. Since some functions expect fields to be of a certain length (e.g., those relating to metabolites, genes, and reactions), empty fields may cause errors in those functions.

This PR implements a change in the `importYaml` function that checks if any model fields are still empty at the end of the import process, and removes those fields. A few exceptions are included to avoid removing some fields which are OK to be empty (`description`, `version`).

This PR also addresses an [Issue](https://github.com/SysBioChalmers/Mouse-GEM/issues/17) in the Mouse-GEM repository.

### DIFF
--- a/code/io/importYaml.m
+++ b/code/io/importYaml.m
@@ -303,6 +303,17 @@ equations(irrevInd) = strcat(leftEqns(irrevInd), ' =>', rightEqns(irrevInd));
 [~, metIdx] = ismember(model.mets, newMets);
 model.S = S(metIdx, :);
 
+% remove empty fields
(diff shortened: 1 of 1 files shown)


haowang-bioinfo:
this PR enables `importYaml` to get rid of empty and non-essential fields - it looks fine to me

haowang-bioinfo (approved):

Issue #401, opened 2022-07-11
prepHumanModelForftINIT should take the paths to reactions.tsv and essential tasks as optional params


johan-gson:
### Description of the issue:
The current code is problematic since it uses "which('reactions.tsv')" - if we use for example the mouse model, it will grab the file for Human-GEM and so forth. I'm making an issue so we don't forget to change this.

#### Expected feature/value/output:
Just add these as params.

#### Reproducing these results:

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

JonathanRob:
Another option instead of making them into params is to code their paths relative to the location of the `prepHumanModelForftINIT` function. Since the location of the files relative to the function is not likely to change often or at all, I think it would be sufficiently stable.

johan-gson:
This would be an improvement to the default option, agreed, good suggestion! I think however that it would still be problematic for the mouse model. So, when using the mouse model, we would still use the function in Human-GEM, but with different files and a different model. The automatic code will then take the reactions.tsv from Human-GEM. In practice, reactions.tsv is most likely the same, but it still doesn't feel right to use the one in Human-GEM. I propose then to use the code you proposed above as default option, but still make possible to send the path in as a parameter for the two files. In practice, this will not matter much, but maybe they will diverge at some point.

johan-gson:
I think we can close this now?

PR #422, opened 2022-10-12
feat: prepHumanModelForftINIT with paths

johan-gson:
This PR fixes issue #401 and now allows for specifying the paths to the tasks and reactions.tsv. This is good when using the code for for example Mouse-GEM.

prepHumanModelForftINIT supports other models such as Mouse-GEM - the paths to relevant files could not be specified earlier.

### DIFF
--- a/code/tINIT/prepHumanModelForftINIT.m
+++ b/code/tINIT/prepHumanModelForftINIT.m
@@ -11,7 +11,7 @@ function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essenti
 %   essentialTasksFilePath  [optional] Path to the metabolicTasks_Essential.txt
 %                           file. Default will use the file from Human-GEM, so
 %                           if using for example Mouse-GEM, supply another path.
-%   rxnsFilePath            [optional] Path to the rxns.tsv
+%   rxnsFilePath            Path to the reactions.tsv
 %                           file. Default will use the file from Human-GEM, so
 %                           if using for example Mouse-GEM, supply another path.
 %
--- a/code/tINIT/prepHumanModelForftINIT.m
+++ b/code/tINIT/prepHumanModelForftINIT.m
@@ -8,7 +8,7 @@ function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essenti
 %                           such as Mouse-GEM.
 %   convertGenes            True if the genes should be converted to symbols 
 %                           ('CD8A' etc., and not ENSEMBL)
-%   essentialTasksFilePath  [optional] Path to the metabolicTasks_Essential.txt
+%   essentialTasksFilePath  Path to the metabolicTasks_Essential.txt
 %                           file. Default will use the file from Human-GEM, so
 %                           if using for example Mouse-GEM, supply another path.
 %   rxnsFilePath            Path to the reactions.tsv
--- a/code/tINIT/prepHumanModelForftINIT.m
+++ b/code/tINIT/prepHumanModelForftINIT.m
@@ -9,11 +9,13 @@ function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essenti
 %   convertGenes            True if the genes should be converted to symbols 
 %                           ('CD8A' etc., and not ENSEMBL)
 %   essentialTasksFilePath  Path to the metabolicTasks_Essential.txt
-%                           file. Default will use the file from Human-GEM, so
-%                           if using for example Mouse-GEM, supply another path.
+%                           file. This file can normally be found under
+%                           HumanGEM/data/metabolicTasks/metabolicTasks_Essential.txt
+%                           If using for example Mouse-GEM, supply another path.
 %   rxnsFilePath            Path to the reactions.tsv
-%                           file. Default will use the file from Human-GEM, so
-%                           if using for example Mouse-GEM, supply another path.
+%                           file. This file can normally be found under
+%                           HumanGEM/model/reactions.tsv
+%                           If using for example Mouse-GEM, supply another path.
 %
 %   Usage: prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essentialTasksFilePath, rxnsFilePath)
 %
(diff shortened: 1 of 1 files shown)


johan-gson:
@haowang-bioinfo @mihai-sysbio Could someone of you review this so I can merge it? It is a small change that makes it possible to run ftINIT properly for Mouse-GEM and similar GEMs.

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -1,22 +1,37 @@
-function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes)
+function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essentialTasksFilePath, rxnsFilePath)
```
mihai-sysbio, line 1: With the more generalized scope, I would forego the `Human` word in the function name.

```suggestion
function prepDataHumanGEM = prepHumanModelForftINIT(model, convertGenes, essentialTasksFilePath, rxnsFilePath)
```

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -1,22 +1,37 @@
...
 % prepHumanModelForftINIT
 %   Generates a "standard" prepData structure from Human-GEM to be used in ftINIT.
 %   This function typically takes a couple of hours to run, so we recommend to only 
 %   run this once and save the results to a .mat file.
 %
+%   model                   Either Human-GEM or a GEM derived from Human-GEM,
+%                           such as Mouse-GEM.
 %   convertGenes            True if the genes should be converted to symbols 
 %                           ('CD8A' etc., and not ENSEMBL)
+%   essentialTasksFilePath  [optional] Path to the metabolicTasks_Essential.txt
```
mihai-sysbio, line 11:
```suggestion
%   essentialTasksFilePath  Path to the metabolicTasks_Essential.txt
```

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -1,22 +1,37 @@
...
+%                           file. Default will use the file from Human-GEM, so
+%                           if using for example Mouse-GEM, supply another path.
+%   rxnsFilePath            [optional] Path to the rxns.tsv
```
mihai-sysbio, line 14:
```suggestion
%   rxnsFilePath            Path to the reactions.tsv
```

code/tINIT/prepHumanModelForftINIT.m
```diff
@@ -1,22 +1,37 @@
...
 
-
+if nargin < 4
+    [ST, I] = dbstack('-completenames');
+    path = fileparts(ST(I).file);
+    rxnsFilePath = fullfile(path,'..','..','model','reactions.tsv');
+    if nargin < 3
+        essentialTasksFilePath = fullfile(path,'..','..','data','metabolicTasks','metabolicTasks_Essential.txt');
+    end
+end
```
mihai-sysbio, line 28: These lines of code would not be needed if no defaults are maintained:
```suggestion
```

mihai-sysbio (commented):
Very nice improvement with explicit paths!

In this case,, I find providing default values for the function more a maintenance cost than a benefit for the user. These functions are not meant to be typed that many times in an analysis, so being explicit about what files are used is a small overhead for a user but less code for maintainers to worry about. For example, if the name or path of the metabolic tasks file changes, as a maintainer, one has to remember to find (at least) this occurrence and update it as well to keep the code functional. I've made some inline suggestions accordingly.

johan-gson:
Hi Mihail Although you have a point, I request that we keep it as it is :) The main reason is that we otherwise break the interface, which for example will stop the code we just submitted for the single-cell modeling paper from working. I also have the argument that I don't think all users will know what a task list is, where to find it etc., and that having to supply these params makes using the function unnecessarily difficult - the first time I used tINIT I didn't know what the task file was and where to find it. The whole code around ftINIT in Human-GEM has that only purpose, to make it easier to use ftINIT - the code is not really needed in that sense, it is just a fairly simple wrapper that removes some normally unwanted reactions and fills in some standard values, such as the path to the task file. The reason for this fix is that I didn't think of the case with the mouse model, and that is why these params are needed.

mihai-sysbio:
> The main reason is that we otherwise break the interface, which for example will stop the code we just submitted for the single-cell modeling paper from working.

This PR is to the develop branch, which I hope was not referenced in the submission. So it would only become problematic if it gets to main. Another option would be to keep this PR on hold for a few months.

> I also have the argument that I don't think all users will know what a task list is, where to find it etc., and that having to supply these params makes using the function unnecessarily difficult - the first time I used tINIT I didn't know what the task file was and where to find it. 

This is an important concern. My approach would be to solve this with improved documentation, by continuing to maintain the Human-GEM-guide.

> The whole code around ftINIT in Human-GEM has that only purpose, to make it easier to use ftINIT - the code is not really needed in that sense, it is just a fairly simple wrapper that removes some normally unwanted reactions and fills in some standard values, such as the path to the task file.

We are on the same page here. My approach is not to simplify things towards a magical one command that does everything, but to be very clear about what each command will do.

> The reason for this fix is that I didn't think of the case with the mouse model, and that is why these params are needed.

Now that this branch exists, it can be used even without merging.

johan-gson:
I think you have a point, when I think of it the main concern would be that there is a substantial risk that people now will use the mouse model with another reactions.tsv.

mihai-sysbio:
For the Mouse model, when using the Mouse-GEM reactions.tsv, should people be running `prepHumanModelForftINIT` on Human-GEM or on Mouse-GEM?

And should consider adding an ftINIT page for how to use it with other models, say Mouse-GEM?

johan-gson:
So, they should use it with Mouse-GEM, and with that reactions.tsv, although it probably doesn't matter. But using the right model is important.

code/tINIT/prepHumanModelForftINIT.m
mihai-sysbio: I am also fine with leaving this as they are now, and postpone any renaming here.

mihai-sysbio (approved):
@johan-gson please merge when you think is the right time. And a reminder - the `Human-GEM-guide` should probably be updated when this makes it in a release.

Issue #136, opened 2019-10-04
Create an "install" function for easy repo installation


JonathanRob:
### Description of the issue:
Currently, there is no automated **"install"** function for the Human-GEM repo, and instead relies on users to manually add the directories to their MATLAB path. Creating an explicit install function to automatically add the (relevant) directories to the user's path would simplify the use of HumanGEM.

Since the repo contains many functions, scripts, datafiles, etc. that were only used once for implementing a curation/fix/update, this install function should not add directories containing such files (e.g., ComplementaryScripts/modelCuration). In doing so, we should also ensure that there are no important functions in these directories, and move them somewhere else if they are.

#### Expected feature/value/output:
A simple install function, e.g., `installHumanGEM.m` placed in the parent directory of the repository that users can run after downloading/cloning the repository and automatically add the relevant directories to their MATLAB path.

#### Current feature/value/output:
Currently, the user must manually add HumanGEM directories to their path.

**I hereby confirm that I have:**
- [X] Checked that a similar issue does not exist already

mihai-sysbio:
@JonathanRob isn't this achieved through https://github.com/SysBioChalmers/Human-GEM-guide ?

JonathanRob:
Yes, and I think now that we have deprecated the curation scripts/data, then my original concern and rationale for having this install function is largely resolved.

PR #423, opened 2022-10-12
feat: add an install script like in GECKO3

johan-gson:
Added an install script:
To install Human-GEM:
cd to root of Human-GEM
HumanGEMInstaller.install
to uninstall:
HumanGEMInstaller.uninstall

What the installation does is to setup the MATLAB path (and remove Human-GEM from it at uninstall). This is very useful when you have multiple versions of Human-GEM on your disk.


JonathanRob:
Finally, an actual solution to #136! Thanks @johan-gson

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+            splitPaths = splitPaths(1,1:end-1);
+            matches = regexp(splitPaths, filter_, 'match');
+            okPaths = cellfun(@isempty, matches);
+            pathsLeft = splitPaths(1,okPaths);
+            newPaths = strcat(char(join(pathsLeft, pathSep)), pathSep);
+        end
+    end
+    
+    properties (Constant)
+      FILE_FILTER = '.*\.git.*'; %should we add ".deprecated" here?
```
mihai-sysbio, line 48: A more general filter would be to ignore all files and folder that start with `.`. I would also ignore everything ending in `.md`.

One way is to use a recursive negative filter, like done here. Another would be to specify the folders which may be recursively added.

johan-gson: So, I prefer this one in general, because it will just work if you add another folder that should be added (which I'm thinking is the more common case?). I don't think it matters super much which. But changing the filter to "." sounds like a good idea, I will experiment with this.

johan-gson: Hmm, maybe you are right, there are many other folders such as data etc. that don't need to be added. But I don't think it is a big thing - this solves the main issue. I did remove the .deprecated - it was a bit complicated to filter on just */.*, since it gets a bit complicated with slashes on Windows and Linux (I cannot match on only ".", then everything containing a dot will be lost), and I had no way of testing both. So, I left it like this. Another future improvement would be to check if there is another Human-GEM installed and warn if there is, that is probably more important. But I'm thinking we could leave it like this? I have many things to attend to these days :)

haowang-bioinfo (approved):

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+% HumanGEMInstaller
+%   Support for installing and uninstalling
+%   Run HumanGEMInstaller.install to install (will set up the path in MATLAB)
+%   Run HumanGEMInstaller.uninstall to clear the path from MATLAB
+%   To install, you first need to cd to the HumanGEM root.
+
+    methods (Static)
+        function install
+            sourceDir = fileparts(which(mfilename));
+            paths = HumanGEMInstaller.GetFilteredSubPaths(sourceDir, HumanGEMInstaller.FILE_FILTER);
```
mihai-sysbio, line 11:
```suggestion
            paths = HumanGEMInstaller.GetFilteredSubPaths(sourceDir);
```

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+            addpath(paths);
+            savepath;
+        end
+        function uninstall
+            sourceDir = fileparts(which(mfilename));
+            paths = HumanGEMInstaller.GetFilteredSubPaths(sourceDir, HumanGEMInstaller.FILE_FILTER);
```
mihai-sysbio, line 17:
```suggestion
            paths = HumanGEMInstaller.GetFilteredSubPaths(sourceDir);
```

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+        end
+        function path = getHumanGEMMainPath()
+			path = fileparts(which(mfilename));
+			path = strrep(path, '\', '/'); %get rid of backslashes in Windows
+			if ~endsWith(path, '/')
+				path = strcat(path,'/');
+			end
+		end
+
+        function newPaths = GetFilteredSubPaths(path_, filter_)
```
mihai-sysbio, line 29:
```suggestion
        function newPaths = GetFilteredSubPaths(path_)
```

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+				error('The path in which Human-GEM resides may not contain path separator chars such as semicolon for this installation to work!');
+			end
+            paths = genpath(path_);
+            splitPaths = strsplit(paths, pathSep);
+            %remove the last, it is empty
+            splitPaths = splitPaths(1,1:end-1);
+            matches = regexp(splitPaths, filter_, 'match');
+            okPaths = cellfun(@isempty, matches);
+            pathsLeft = splitPaths(1,okPaths);
+            newPaths = strcat(char(join(pathsLeft, pathSep)), pathSep);
```
mihai-sysbio, line 43:
```suggestion
            subpaths = arrayfun(@(subf) sprintf('%s', [path_ filesep subf{:}]), HumanGEMInstaller.SUBFOLDERS, 'UniformOutput', false);
            subpaths = [strjoin(subpaths, pathsep) pathsep];
            newPaths = subpaths;
```

HumanGEMInstaller.m
```diff
@@ -0,0 +1,50 @@
...
+        end
+    end
+    
+    properties (Constant)
+      FILE_FILTER = '(.*\.git.*)|(.*\.deprecated.*)';
```
mihai-sysbio, line 48:
```suggestion
      SUBFOLDERS = {'code', 'data', 'model'};
```

mihai-sysbio (commented):
With the suggestion mode I've explored a different approach of only adding 3 statically-defined top-level folders, thus avoiding to add any "wrong" folders, and having to filter through the generated subpaths of those folders.

johan-gson:
Hi Mihail, this code doesn't work for me, you are not adding the subfolders in code and so forth (at least not on my computer). Could you either fix that (you could use my previous strategy for the code subtree) or revert this? I simply don't have time to dig further into this now, but I think it is important that it works.

johan-gson:
Or, was there something here I didn't get? Are you saying that we don't need to add all the subfolders? I don't think MATLAB will find the code otherwise, right?

johan-gson:
And, I realize I technically committed the code :). Still, I don't know if you want to work more or go back :)

mihai-sysbio:
> And, I realize I technically committed the code :). Still, I don't know if you want to work more or go back :)

It's fine, I'd like to continue fixing the obvious `genpath` that I forgot to add. For some reason I'm having VPN issues so Matlab won't start.

HumanGEMInstaller.m
```diff
@@ -0,0 +1,45 @@
...
+        function newPaths = GetFilteredSubPaths(path_)
+            pathSep = pathsep();
+			%Check that there are no separators in the path - that will cause 
+            %problems since the separator is used to separate paths in a string
+			if contains(path_, pathSep)
+				error('The path in which Human-GEM resides may not contain path separator chars such as semicolon for this installation to work!');
+			end
+            subpaths = arrayfun(@(subf) sprintf('%s', [path_ filesep subf{:}]), HumanGEMInstaller.SUBFOLDERS, 'UniformOutput', false);
+            subpaths = [strjoin(subpaths, pathsep) pathsep];
+            newPaths = subpaths;
```
mihai-sysbio, line 38: Here I'm adding the `genpath` that I missed in the previous suggestions:

```suggestion
            subpaths = arrayfun(@(subf) sprintf('%s', genpath([path_ filesep subf{:}])), HumanGEMInstaller.SUBFOLDERS, 'UniformOutput', false);
            newPaths = strjoin(subpaths, '');
```

mihai-sysbio (changes requested):

HumanGEMInstaller.m
mihai-sysbio: And now I realised:
- `HumanGEMInstaller.m` is outside the `code` folder, thus not in the path; one idea is to move it there
- `Human-GEM-guide` should probably also be updated when this is merged

johan-gson:
Looks good, it seems to work now!

mihai-sysbio:
> Looks good, it seems to work now!

Also when changing directory and trying to uninstall? I tried to mention [before](https://github.com/SysBioChalmers/Human-GEM/pull/423#discussion_r996568059) that this check fails unless this script is moved to the `code` directory, which is in the path, whereas the main `Human-GEM` directory is not.

johan-gson:
So, I added the Human-GEM folder to the path - I think it should be in there. In my opinion, the install script should be there, otherwise people will not find it. I tested, it works.

mihai-sysbio:
> the install script should be there

I can see arguments on both sides to having the install script in the top level folder or in `/code`. Perhaps we should ask for other opinions @JonathanRob @haowang-bioinfo ?

haowang-bioinfo:
would be neat to place it in `/code`, while a dedicated section explaining install/uninstall with this script should be added to the top level `README`

JonathanRob:
I'm ok with the script being in either location. Normally such things would of course be in `/code`, but an install script is a special case where I can see the benefit of keeping it in a more obvious location.

johan-gson:
For me it doesn't matter that much as long as it works. If someone wants to move it, please go ahead!

code/HumanGEMInstaller.m
```diff
@@ -0,0 +1,44 @@
...
+
+        function newPaths = GetFilteredSubPaths(path_)
+            pathSep = pathsep();
+			%Check that there are no separators in the path - that will cause 
+            %problems since the separator is used to separate paths in a string
+			if contains(path_, pathSep)
+				error('The path in which Human-GEM resides may not contain path separator chars such as semicolon for this installation to work!');
+			end
+            subpaths = arrayfun(@(subf) sprintf('%s', genpath([path_ filesep subf{:}])), HumanGEMInstaller.SUBFOLDERS, 'UniformOutput', false);
+            newPaths = [strjoin(subpaths, '') pathSep path_];
```
mihai-sysbio, line 37: The installer file was moved and documented in [2e3f01c](https://github.com/SysBioChalmers/Human-GEM/pull/423/commits/2e3f01cf6ac345a0909f53c674fd80beea9b3764). The reason to go ahead was that we need to be mindful that users might not use Matlab, or care about the `code` in this repository at all, so a clean top-level folder is more appealing to them. Now, the previous fix is not longer required and could be reverted:
```suggestion
            newPaths = strjoin(subpaths, '');
```

mihai-sysbio (approved):

johan-gson:
Looks good, good job Mihail. I'm merging this now!

Issue #434, opened 2022-11-17
Issues with the HMDB IDs of metabolites


orsiliska:
### Description of the issue:
There are two issues with the HMDB IDs of the metabolites in the model.
1: One metabolite, 	MAM03055r (trioxilin A3) has a CHEBI ID annotated as its HMDB ID (even though the metabolite is included in HMDB as well: HMDB0001977).

2: The HMDB IDs used in the model don't all come from the updated version of HMDB. HMDB made its IDs longer with version 4 (the current versions is 5). In many cases the model contains the old IDs, which are formatted as "HMDB" followed by a 5-digit number. In the newer version this was changed to "HMDB" followed by a 7-digit number. Currently this mix of versions makes it hard to consistently connect to other datasets using HMDB IDs. 

#### Expected feature/value/output:
HMDB IDs should be fixed to all be in the format of "HMDB" followed by a 7-digit number. This shouldn't be a problem, as you'd  simply need to insert two 0s in front of the 5-digit number in the ID for metabolites that are still annotated with the old version.

```

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [ ] Done this analysis in the `main` branch of the repository
- [X ] Checked that a similar issue does not exist already

haowang-bioinfo:
@orsiliska please review the [changes](https://github.com/SysBioChalmers/Human-GEM/pull/435/files) made according to your suggestions

haowang-bioinfo:
fixed in #435

PR #435, opened 2022-11-18
fix: update HMDB IDs to version 4 

haowang-bioinfo:
* update HMDB IDs to version 4 format by inserting double zero, and fix HMDB ID for MAM03055r, as suggested in #434


orsiliska (changes requested):
Thank you for the quick response!

I looked over the updated IDs. They are mostly correct, with the exception of 23 IDs.
I'm attaching a table of the incorrect IDs and their correct versions.

In short:
 - 5 IDs have now been revoked by HMDB as the records weren't reliable.
 - 9 IDs have been replaced by new ones. The updated version of these IDs can still be found in HMDB as secondary IDs.
 - 8 IDs were updated when they didn't need to be.
 - 1 ID, that of D-aspartate originally contained a typo, but didn't need to be updated otherwise.

All other updated IDs are currently primary IDs in HMDB, and should be kept kept in the model in their updated form.
[hmdbs_to_fix.txt](https://github.com/SysBioChalmers/Human-GEM/files/10041906/hmdbs_to_fix.txt)

haowang-bioinfo:
@orsiliska excellent!

The 23 IDs reported in above table had been corrected and are ready for your inspection.

feiranl (approved):
LGTM!

PR #437, opened 2022-12-01
feat: add compartment information to genes.tsv

haowang-bioinfo:
* compartment information is added as discussed in #436
  - subcellular locations of metabolic genes were extracted from experimentally validated [Cell Atlas](https://www.proteinatlas.org/search/has_protein_data_in%3ACell) and manually curated [SwissProt](https://www.uniprot.org/) data sources
  - compartment info from both sources were combined and integrated into `genes.tsv` with two columns: `compartments` and `compDataSource`


feiranl:
I will add the predicted compartment soon to this file.

haowang-bioinfo:
> I will add the predicted compartment soon to this file.

maybe better to add prediction in another PR after this

code/modelCuration/getCompFromUniprotCellAtlas.py
```diff
@@ -0,0 +1,138 @@
...
+    if pd.isna(geneComps_from_swissprot[i]) and geneComps_from_cell_atlas[gene] != []:
+        out_list = geneComps_from_cell_atlas[gene]
+        source[i] = 'CellAtlas'
+    elif not pd.isna(geneComps_from_swissprot[i]) and geneComps_from_cell_atlas[gene] == []:
+        out_list = geneComps_from_swissprot[i].split(';')
+        source[i] = 'SwissProt'
+    elif not pd.isna(geneComps_from_swissprot[i]) and geneComps_from_cell_atlas[gene] != []:
+        union = set(geneComps_from_swissprot[i].split(';')+ geneComps_from_cell_atlas[gene])
+        out_list = list(union)
+        source[i] = 'SwissProt&CellAtlas'
```
JonathanRob, line 127: For cases when `source` is more than one, maybe use `;` as the delimiter instead of `&`, since we typically use the former as a delimiter.

code/modelCuration/getCompFromUniprotCellAtlas.py
```diff
@@ -0,0 +1,138 @@
...
+    if not pd.isna(text):
+        for key in swissprot_keywords:
+            # check if cellular location text contains any keyword from a compartment
+            if any(word in text.lower() for word in swissprot_keywords[key]):
+                out_list.append(key)
+        comps_from_swissprot[SwissProt_proteins[i]] = out_list
+
+
+# load Human-GEM gene annotation file
+HumanGenes_tsv = pd.read_table("~/Documents/GitHub/Human-GEM/model/genes.tsv")
```
mihai-sysbio, line 41: I would suggest a relative path here.

mihai-sysbio (changes requested):
Having the localization source as a separate column in this file can get tricky. For example, when there are 3 localizations and 2 sources:

genes | compartments | compDataSource
-- | -- | --
ENSG00000003987 | Golgi apparatus;Cytosol;Nucleus | SwissProt&CellAtlas

This might mean agreement between sources, or it might mean that it is a "sum". To avoid this source of confusion, I suggest the removal of this column from `genes.tsv`. If the exact source is needed, one should look directly into the data files.

code/modelCuration/getCompFromUniprotCellAtlas.py
```diff
@@ -0,0 +1,138 @@
...
+Human_genes = HumanGenes_tsv['genes'].to_list()
+Human_proteins = HumanGenes_tsv['geneUniProtID'].to_list()
+
+
+# save SwissProt compartment information to tsv file
+swissprot_compList = ['']*len(Human_genes)
+for i in range(len(Human_genes)):
+    if not pd.isna(Human_proteins[i]):
+        if Human_proteins[i] in SwissProt_proteins:
+            swissprot_compList[i] = ';'.join(comps_from_swissprot[Human_proteins[i]])
```
feiranl, line 51: Is that possible also extract the info of wether the protein is a membrane protein?

feiranl (approved):
👍

code/modelCuration/getCompFromUniprotCellAtlas.py
haowang-bioinfo: yes, there's curated info about membrane proteion

code/modelCuration/getCompFromUniprotCellAtlas.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+# integrate compartment info from cellAtlas and SwissProt with following rules:
+# 1. output two columns: "compartments" and "compDataSource"
+# 2. use one source if another has no assigned compartments
+# 3. union compartments if both source are provided
+swissprot_comps = pd.read_table("../../data/modelCuration/Swissprot_compartments.tsv")
+geneComps_from_swissprot = swissprot_comps['compartments'].to_list()
+
+# combine compartment info from two data sources
```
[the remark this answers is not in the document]
haowang-bioinfo: after rechecking the original table, the comp info was determined by the reliability of either IH or IF. While in the uploaded table, I only provided IF reliability that's a bit bias

PR #440, opened 2022-12-15
feat: add compartment info from deeploc2 prediction

feiranl:
* compartment information is added as discussed in #436
  - subcellular locations of metabolic genes were predicted using [DeepLoc2]([10.1093/nar/gkac278](https://doi.org/10.1093/nar/gkac278)) using the default parameter
  - compartment info for genes which doesn't have localization annotation as in pull request #437 were combined and integrated into `genes.tsv` with `compDataSource` was denoted as `DeepLoc2`

*Note: replace [ ] with [X] to check the box. PLEASE DELETE THIS LINE*


haowang-bioinfo:
very good!

feiranl:
Then those corresponding rxns should involve both extracellular and cytosol metabolites?

haowang-bioinfo:
> Then those corresponding rxns should involve both extracellular and cytosol metabolites?

do you mean the enzymes are potential transporters?

haowang-bioinfo (approved):

feiranl:
> > Then those corresponding rxns should involve both extracellular and cytosol metabolites?
> 
> do you mean the enzymes are potential transporters?

Could be but not necessarily. It could functions as to degrade extracellular metabolites into extracellular molecules that can be transported into cytosol, or just perform cytosol conversions. So I say when the protein is annotated to cell membrane, so either cytosol or extracellular compartment for metabolites in the corresponding rxns should be ok.

Issue #444, opened 2023-01-05
remove non-transporter genes from transport reactions


haowang-bioinfo:
### Description of the issue:
* A total of **20** non-transporter genes are discovered in the GPR rules of following `Transport reactions`:

| gene            | reasons                                                                                                        | subsystems                                                                        | rxnIDs                                                                                                                                                                                                                                    |
|-----------------|-----------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| ENSG00000008300 | annotated as signalling receptor in Uniprot                                                                     | Transport reactions                                                               | MAR06525                                                                                                                                                                                                                                  |
| ENSG00000026103 | annotated as tumor necrosis receptor mediating apoptosis | Transport reactions                                                               | MAR00030                                                                                                                                                                                                                                  |
| ENSG00000101255 | Inactive protein kinase                                                                                         | Transport reactions                                                               | MAR00025                                                                                                                                                                                                                                  |
| ENSG00000101558 | not related to fatty acid transport                                                                             | Transport reactions                                                               | MAR00730                                                                                                                                                                                                                                  |
| ~~ENSG00000101892~~ | ~~Has lost its ancestral function as a Na,K-ATPase beta-subunit~~                                                   | ~~Transport reactions~~                                                               | ~~MAR05295;MAR05429~~                                                                                                                                                                                                                         |
| ENSG00000108468 | cannot determine wether the rxn is related to the annotation, because it's annotated as transcription regulator | Transport reactions                                                               | MAR00030                                                                                                                                                                                                                                  |
| ENSG00000110243 | inconsistent annotation with rxn                                                                                | Transport reactions                                                               | MAR00029;MAR00030                                                                                                                                                                                                                         |
| ENSG00000112981 | Does not seem to have NDK kinase activity.                                                                      | Nucleotide metabolism;Purine metabolism;Pyrimidine metabolism;Transport reactions | MAR04016;MAR04018;MAR04044;MAR04291;MAR04420;MAR04570;MAR04572;MAR04006;MAR04028;MAR04030;MAR04670;MAR04673;MAR04675;MAR06614;MAR06615;MAR06616;MAR04635;MAR07878;MAR07883;MAR07886;MAR07890;MAR07894;MAR07895;MAR04908;MAR04907;MAR04906 |
[...]

Issue #426, opened 2022-10-22
Removal of MEIKIN from fatty acid activation reactions


haowang-bioinfo:
### Description of the issue:
* [MEIKIN](https://www.uniprot.org/uniprotkb/A0A087WXM9/entry) ([ENSG00000239642](https://www.ensembl.org/Homo_sapiens/Gene/Summary?g=ENSG00000239642)) is a [key regulator of kinetochore function](https://pubmed.ncbi.nlm.nih.gov/25533956/) during meiosis I. Specifically, it regulate kinases in [cohesin protection](https://pubmed.ncbi.nlm.nih.gov/33888556/). However, it was mis-annotated as one of isoenzymes that catalyze [fatty acid activation reactions](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000239642) across compartments of cytosol, mitochondria, peroxisome, and ER.

#### Current feature/value/output:
* The meiosis regulator MEIKIN (ENSG00000239642) is now involved into a total of 132 fatty acid activation reactions

#### Expected feature/value/output:
* Remove MEIKIN (ENSG00000239642) gene from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

haowang-bioinfo:
With the discovery of this (ENSG00000239642) gene, it can thus be speculated that there might be others that are actually not involve into metabolic process but were incorporated HumanGEM, probably due to sequence similarity.

A systematic investigation might be necessary at some point. The [GPT](https://openai.com/blog/chatgpt/) language model appears to be useful tools undertaking this task.

haowang-bioinfo:
As a follow-up of above comment, the Uniprot annotation of Human-GEM genes had been screened by GPT-3 model for preliminary evaluation, the obtained non-metabolic candidates were then subjected to manual check. The results are now reported as issues #443, #444, #446, #447, and more for open discussion, before subsequent implementations in pull requests.

mihai-sysbio:
> As a follow-up, above comment triggered `GPT3 guided curation`. First the Uniprot annotation of Human-GEM genes had been screened by GPT-3 model for preliminary evaluation, the obtained non-metabolic candidates were then subjected to manual check.

Nice description! Is this process reproducible? I'm thinking that evidence evolves, so it might be worth it to do this process again in the future. Also, are there any simulation checks ran after these deletions?

[...]

PR #450, opened 2023-01-10
fix-gene: remove non-transporter genes - updated

haowang-bioinfo:
* remove **20** non-transporter genes from `Transporter reactions`, as proposed in #444


mihai-sysbio:
See my comment in [#444 ](https://github.com/SysBioChalmers/Human-GEM/issues/444#issuecomment-1378199203) - I'm worried that we might be moving too fast here.

johan-gson (approved):
It was difficult to see the exact changes from last time, but I think you thought it through - I will not go over everything again. Good job!

haowang-bioinfo:
Simulation checks with models before and after these deletions showed exactly the same growth rates.

JonathanRob:
> simulation checks with models before and after these deletions were conducted and showed exactly the same growth rates

Why would removing genes change simulation results? Unless you're using an enzyme-constrained model, I don't quite understand how it would influence flux at all (assuming by "simulation" you mean FBA).

haowang-bioinfo:
> Why would removing genes change simulation results? Unless you're using an enzyme-constrained model, I don't quite understand how it would influence flux at all (assuming by "simulation" you mean FBA).

yes, now it's just FBA that gave the same results since stoichiometric was not changed. As just commented in #426, a systematic checking code would be needed as reaction addition/removal may happen in the future.

feiranl:
Gene essentiality analysis should performed in terms of removing genes? The check function is coming.

haowang-bioinfo:
> Gene essentiality analysis should performed in terms of removing genes?

agree

johan-gson:
If you want I can do the gene essentiality part - I recently did this for the single-cell modeling paper, so I know how to do it and have the code on the cluster etc. set up for this. I can do it in 1-2 weeks (or later), I don't know the status of the model, when it will be ready.

johan-gson:
But don't expect any large changes to the prediction power :)

Issue #447, opened 2023-01-09
Remove ST20 from Pentose phosphate pathway


haowang-bioinfo:
### Description of the issue:
* [ST20](https://www.uniprot.org/uniprotkb/Q9HBF5) (ENSG00000180953) is a short peptide that may act as a [tumor suppressor promoting apoptosis of cancer cells](https://pubmed.ncbi.nlm.nih.gov/11857354/).

#### Current feature/value/output:
* ST20 is now assigned as an isoenzyme of phosphopentomutase in [MAR04710](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR04710).

#### Expected feature/value/output:
* Remove this gene from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

haowang-bioinfo:
resolved in #451

PR #451, opened 2023-01-13
fix: remove ST20 from PPP pathway

haowang-bioinfo:
* remove ST20 (ENSG00000180953) from PPP pathway, as proposed in #447


JonathanRob (approved):

johan-gson (approved):
Looks good to me. Checked the gene using ChatGPT, it said it is not known what it is good for, more than a growth regulation in cancer. So I would not think it is a metabolic enzyme.

Issue #452, opened 2023-01-13
remove non-metabolic genes from Inositol phosphate metabolism


haowang-bioinfo:
### Description of the issue:
* Seven non-metabolic genes are discovered in the GPR rules of `Inositol phosphate metabolism`:

| gene            | Reasons                                                                                  | Subsystems                    | RxnIDs                              |
|-----------------|------------------------------------------------------------------------------------------|-------------------------------|-------------------------------------|
| [ENSG00000115896](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000115896) | has no PLC activity to phosphatidylinositol 4,5-bisphosphate and phosphatidylinositol  | Inositol phosphate metabolism | MAR00663;MAR06559;MAR08822          |
| [ENSG00000117598](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000117598) | has most probably no lipid phosphatase activity                         | Inositol phosphate metabolism | MAR06545;MAR06547;MAR06557;MAR06591 |
| [ENSG00000126264](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000126264) | the annotation doesn't match the rxn                                         | Inositol phosphate metabolism | MAR06555                            |
| [ENSG00000154822](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000154822) | annotated as inactive                                                                    | Inositol phosphate metabolism | MAR00663;MAR06559;MAR08822          |
| [ENSG00000158079](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000158079) | no enough supporting evidence                                                            | Inositol phosphate metabolism | MAR06545;MAR06547;MAR06557;MAR06591 |
| [ENSG00000160539](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000160539) | inactive enzyme in KEGG and Uniprot | Inositol phosphate metabolism | MAR06545;MAR06547;MAR06557;MAR06591 |
| [ENSG00000276231](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000276231) | annotated as kinase regulator                                                            | Inositol phosphate metabolism | MAR06555                            |

#### Current feature/value/output:
* The association of these genes is lack of evidence according to the provided reasons

#### Expected feature/value/output:
* Remove above genes from Human-GEM 

[...]

PR #455, opened 2023-01-17
fix: remove non-metabolic genes from Inositol phosphate metabolism

haowang-bioinfo:
* remove **7** non-metabolic genes, as proposed in #452


johan-gson (approved):
Checked all of the changes, you did what you said in the issue, so looks good to me!

Issue #466, opened 2023-01-27
remove non-metabolic gene from Cysteine and methionine metabolism


feiranl:
### Description of the issue:
* non-metabolic gene is discovered in the GPR rules of `Cysteine and methionine metabolism`:

| gene            | Reasons                                                                                  | Subsystems                    | RxnIDs                              |
|-----------------|------------------------------------------------------------------------------------------|-------------------------------|-------------------------------------|
| [ENSG00000147485](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000147485) | [Probable oxidoreductase (Probable). Lacks peroxidase activity](https://www.uniprot.org/uniprotkb/A1KZ92/entry) | Cysteine and methionine metabolism;Miscellaneous;Phenylalanine, tyrosine and tryptophan biosynthesis;Purine metabolism;Xenobiotics metabolism | MAR06602;MAR06603;MAR06808;MAR07628;MAR06523;MAR06535;MAR06536;MAR04201 |

#### Current feature/value/output:
* The association of these genes is lack of evidence according to the provided reasons

#### Expected feature/value/output:
* Remove above genes from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

Issue #465, opened 2023-01-26
remove non-metabolic genes from Arginine and proline metabolism


feiranl:
### Description of the issue:
* non-metabolic gene is discovered in the GPR rules of `Arginine and proline metabolism`:

| gene            | Reasons                                                                                  | Subsystems                    | RxnIDs                              |
|-----------------|------------------------------------------------------------------------------------------|-------------------------------|-------------------------------------|
| [ENSG00000243709](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000243709) | [reaction function and annotation doesnot match](https://www.uniprot.org/uniprotkb/O75610/entry) | Arginine and proline metabolism | MAR03833;MAR03835;MAR03837;MAR04778;MAR04779;MAR04780;MAR04781;MAR08609 |

#### Current feature/value/output:
* The association of these genes is lack of evidence according to the provided reasons

#### Expected feature/value/output:
* Remove above genes from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

Issue #464, opened 2023-01-26
remove non-metabolic genes from Alanine, aspartate and glutamate metabolism


feiranl:
### Description of the issue:
* three non-metabolic genes are discovered in the GPR rules of `Alanine, aspartate and glutamate metabolism`:

| gene            | Reasons                                                                                  | Subsystems                    | RxnIDs                              |
|-----------------|------------------------------------------------------------------------------------------|-------------------------------|-------------------------------------|
| [ENSG00000137261](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000137261) | [believed to be not involved in catalysis](https://www.uniprot.org/uniprotkb/Q5T4B2/entry) | Alanine, aspartate and glutamate metabolism | MAR04285 |
| [ENSG00000146166](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000146166) | [ Does not seem to have enzymatic activity (By similarity).](https://www.uniprot.org/uniprotkb/Q5TDP6/entry) | Alanine, aspartate and glutamate metabolism | MAR03890 |
| [ENSG00000241258](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000241258) | [DNA-directed RNA polymerase III - completely irrelevant](https://www.uniprot.org/uniprotkb/O75575/entry) | Alanine, aspartate and glutamate metabolism | MAR03813 |

#### Current feature/value/output:
* The association of these genes is lack of evidence according to the provided reasons

#### Expected feature/value/output:
* Remove above genes from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

haowang-bioinfo:
very good, go ahead please

Issue #463, opened 2023-01-26
remove non-metabolic genes from Lysine metabolism


feiranl:
### Description of the issue:
* two non-metabolic genes are discovered in the GPR rules of `Lysine metabolism`:

| gene            | Reasons                                                                                  | Subsystems                    | RxnIDs                              |
|-----------------|------------------------------------------------------------------------------------------|-------------------------------|-------------------------------------|
| [ENSG00000005483](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000005483) | [inactive enzyme from KEGG,](https://www.genome.jp/dbget-bin/www_bget?hsa:55904) [ it does not exhibit histone methyltransferase towards histone H3](https://www.uniprot.org/uniprotkb/Q8IZD2/entry) | Lysine metabolism | MAR04241;MAR06975;MAR06976;MAR06977;MAR08025;MAR08026;MAR08027 |
| [ENSG00000167123](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000167123) | [ Does not express any beta-galactosyltransferase activity in vitro](https://www.uniprot.org/uniprotkb/Q5T4B2/entry) | Lysine metabolism | MAR06983 |

#### Current feature/value/output:
* The association of these genes is lack of evidence according to the provided reasons

#### Expected feature/value/output:
* Remove above genes from Human-GEM 

**I hereby confirm that I have:**
- [X] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

PR #467, opened 2023-01-27
fix: remove non-metabolic genes from amino acid metabolism

feiranl:
* remove **7** non-metabolic genes, as proposed in #463 #464  #465 #466 


johan-gson (approved):
Looks good to me, good work! I have not checked that it is relevant to remove these genes, only that you removed the genes that were named in the issues.

haowang-bioinfo (approved):
looks very good!

Issue #493, opened 2023-03-06
update use of function `children` in `simplifyGrRules`


mihai-sysbio:
#### Current behavior:

A function in Matlab has changed its output, so the use of said function should be revised.

#### Expected feature/value/output:

The `simplifyGrRules` should be updated to be compatible with Matlab's `children` function, whose [behavior has changed since Matlab 2020b](https://la.mathworks.com/help/symbolic/sym.children.html#mw_4a28b8ff-58a4-4aa6-ae74-092938e39e3a). 

#### Reproducing these results:

In Matlab 2021b, running this function brings up errors.

mihai-sysbio:
With the PR merged, this issue may be closed.

PR #494, opened 2023-03-06
fix: adapt simplifyGrRules to updates in Matlab children function

mihai-sysbio:

The branch/commit was created by @JonathanRob, and based on the commit description I believe it aimed to resolve the now-described issue #493.


JonathanRob:
Since I no longer have a Matlab license, I'm unfortunately not able to test this.

mihai-sysbio:
The fix works for me, with the single exception being a Matlab limitation:
> Failed to expand grRule #7584 due to excess length. Rule will remain in simplified form.

Based on this, I'll go ahead and merge this PR.

Issue #510, opened 2023-03-23
feat: remove pseudogenes, read through genes, and ones with only predicted functions


haowang-bioinfo:
### Description of the issue:

It is proposed to remove following genes, because they are either pseudogenes, or genes from read through transcripts (from two adjacent genes in the same strand with same orientation), or the ones with only predicted functions.

|      gene id   |      check results                              |
|----------------|-------------------------------------------------|
| [ENSG00000011052](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000011052)|read through                                     |
| [ENSG00000099984](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000099984)|pseudogene                                       |
| [ENSG00000100101](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000100101)|novel transcript                                 |
| [ENSG00000114786](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000114786)|read through                                     |
| [ENSG00000125954](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000125954)|read through                                     |
| [ENSG00000133475](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000133475)|pseudogene                                       |
| ~~[ENSG00000137700](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000137700)~~|~~with only predicted function~~                |
| [ENSG00000168970](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000168970)| read through                                     |
| [ENSG00000172782](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000172782)| with only predicted function  |
| [ENSG00000189366](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000189366)| pseudogene, only predicted function  |
| [ENSG00000250741](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000250741)| read through, no annotation |
| [ENSG00000255730](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000255730)| with only predicted function |
| [ENSG00000259030](https://metabolicatlas.org/explore/Human-GEM/gem-browser/gene/ENSG00000259030)| read through |
[...]

PR #513, opened 2023-03-25
fix: remove pseudogenes, read through genes, and ones with only predicted functions

haowang-bioinfo:
This PR removes 21 that are pseudogenes, read through genes, and ones with only predicted functions, as proposed in #510.

Note: all these genes are existed as isoenzymes in GPRs, so that their removal has no much impact to the model.


feiranl (approved):
LGTM!

Issue #526, opened 2023-04-05
Changes to GPRs for Cytochrome C Oxidoreductase (AKA Complex IV of the Electron Transport Chain)


Devlin-Moyer:
The GPRs for MAR06914 and MAR13081 (cytochrome c oxidoreductase; complex IV of the electron transport chain) were recently changed from

> ENSG00000198804 and ENSG00000198712 and ENSG00000198938 and ENSG00000131143 and ENSG00000131055 and ENSG00000178741 and ENSG00000135940 and ENSG00000111775 and ENSG00000156885 and ENSG00000126267 and ENSG00000160471 and ENSG00000164919 and ENSG00000161281 and ENSG00000112695 and ENSG00000131174 and ENSG00000170516 and ENSG00000127184 and ENSG00000176340 and ENSG00000187581 and ENSG00000189043

to

> ENSG00000111775 and ENSG00000112695 and ENSG00000126267 and ENSG00000127184 and ENSG00000131143 and ENSG00000131174 and ENSG00000135940 and ENSG00000164919 and ENSG00000176340 and ENSG00000178741 and ENSG00000198712 and ENSG00000198804 and ENSG00000198938 and ENSG00000189043

but several of the genes that were removed can functionally replace one of the genes that remain, and are often expressed in a mutually-exclusive fashion across different tissues. [source](https://doi.org/10.1155/2017/1534056) Specifically:

- COX4I1 (ENSG00000131143) is ubiquitously expressed and COX4I2 (ENSG00000131055) is only expressed in hypoxic conditions 
- COX6A1 (ENSG00000111775) is generally expressed everywhere but muscle tissues and COX6A2 (ENSG00000156885) is generally only expressed in muscle tissues
- COX6B1 (ENSG00000126267) is expressed in somatic tissues and COX6B2 (ENSG00000160471) is only expressed in testes
- COX7A1 (ENSG00000161281) is generally only expressed in muscle tissues and COX7A2 (ENSG00000112695) is generally expressed everywhere but muscle tissues
- COX8A (ENSG00000176340) is ubiquitously expressed and COX8C (ENSG00000187581) is only expressed in some tissues, although its functional role is apparently unknown, so maybe it should stay out of this GPR

~Also, [NDUFA4 (ENSG00000189043) is required for wild-type catalytic activity of Complex IV](https://doi.org/10.1016/j.tem.2018.03.009), but is currently absent from the GPRs of both reactions~

[...]

PR #525, opened 2023-04-05
fix: refine reactions, reversibility, and GPRs of Isocitrate Dehydrogenase metabolism

haowang-bioinfo:
This PR improves GPRs of isocitrate dehydrogenase reactions, as proposed by #523:
- remove MAR04111, MAR04112, MAR04113, MAR04585, MAR04586, MAR04587, MAR04588, because as reported in #524:
  - MAR00710 is equivalent to MAR04111+MAR04585
  - MAR00712 is equivalent to MAR04113+MAR04587
  - MAR04588 is part of MAR03957
  - MAR03958 is equivalent to MAR04112+MAR04586
- change to reversible reactions for MAR00710, MAR00712, and MAR03958
- change GPRs to
  - `ENSG00000138413`: for MAR0710, MAR0712, 
  - `ENSG00000182054`: for MAR03958
  - `ENSG00000067829 and ENSG00000101365 and ENSG00000166411`: for MAR03957
- associate Rhea IDs to MAR0710, MAR0712, MAR03957, MAR03958

**I hereby confirm that I have:**

- [ ] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [ ] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

Issue #524, opened 2023-04-04
Parallel/Redundant Versions Of Reactions


Devlin-Moyer:
There are a number of cases where the "same" overall reaction is represented at multiple different levels of resolution in different reactions/series of reactions:

- [x] The aconitase reaction (citrate <-> isocitrate) can be carried out in
  - one step:MAR04456
  - two steps: MAR04458 (citrate <-> cis-aconitate + H<sub>2</sub>O) and MAR04589 (cis-aconitate + H<sub>2</sub>O <-> isocitrate)
- [x] The cytosolic isocitrate dehydrogenase (IDH) reaction (isocitrate + NADP<sup>+</sup> -> AKG + CO<sub>2</sub> + NADPH) can be carried out in
  - one step: MAR00710
  - two steps: MAR04585 (isocitrate + NADP<sup>+</sup> <-> oxalosuccinate + NADPH) and MAR04111 (oxalosuccinate -> AKG + CO<sub>2</sub>)
- [x] The peroxisomal IDH reaction can be carried out in
  - one step: MAR00712
  - two steps: MAR04587 (isocitrate + NADP<sup>+</sup> <-> oxalosuccinate + NADPH) and MAR04113 (oxalosuccinate -> AKG + CO<sub>2</sub>)
- [x] The NADP(H)-dependent mitochondrial IDH reaction can be carried out in
  - one step: MAR03958
  - two steps: MAR04586 (isocitrate + NADP<sup>+</sup> <-> oxalosuccinate + NADPH) and MAR04112 (oxalosuccinate -> AKG + CO<sub>2</sub>)
- [x] The pyruvate dehydrogenase (PDH) reaction (pyruvate + NAD<sup>+</sup> -> acetyl-CoA + CO<sub>2</sub> + NADH) can be carried out in
  - one step: MAR04137
  - three steps: MAR06410 (pyruvate + lipoamide + H<sup>+</sup> -> S-acetyldihydrolipoamide + CO<sub>2</sub>), MAR06412 (S-acetyldihydrolipoamide + CoA -> acetyl-CoA + dihydrolipoamide), and MAR06409 (lipoamide + NADH + H<sup>+</sup> <-> dihydrolipoamide + NAD<sup>+</sup>)
- [x] The 2-oxoglutarate dehydrogenase (OGDH) reaction (AKG + CoA + NAD<sup>+</sup> -> succinyl-CoA + CO<sub>2</sub> + NADH) can be carried out in
[...]

PR #532, opened 2023-04-14
fix: update GPRs for Cytochrome C Oxidoreductase and remove MAR04456 that is duplicated to MAR04458+MAR04589

haowang-bioinfo:
This PR resolves two issues (this should be avoided):
- fixes GPRs for Cytochrome C Oxidoreductase, as discussed in #526
  - updated GPR of MAR06914 and MAR13081
  - blocked reaction MAR13081, the ROS version Complex IV, by setting ub=lb=0
  - removed COX8C (ENSG00000187581) that has unknown tissue specificity and unclear function
- modification as reported in #524 and discussed in #525
  - remove MAR04456 that is duplicated to MAR04458+MAR04589
  - assign compartment-specific GPRs as:
    - MAR04454: ACO1 (ENSG00000122729) - cytosolic
    - MAR04458 and MAR04589: ACO2 (ENSG00000100412) - mitochondrial
  - remove ENSG00000136381 (IREB2), a RNA-binding protein that binds to iron-responsive elements and repress its translation, from [MAR04454](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR04454) and the model


Devlin-Moyer (approved):

haowang-bioinfo:
@Devlin-Moyer maybe check again this PR after more changes were made

Devlin-Moyer (approved):

Devlin-Moyer (approved):

model/Human-GEM.yml
```diff
@@ -302273,7 +302273,7 @@
           - MAM02630m: -1
           - MAM02631m: 0.02
       - lower_bound: 0
-      - upper_bound: 1000
```
feiranl: Is it alright to leave the bound as it is before? I'm not certain which reaction will carry the flux, whether this ROS one or the regular one. I personally vote for keeping them both open. In the step of specific model generation, the user can remove either one of these options. Blocking them in the generic model could lead to unnecessary issues in the specific model simulation, making it difficult to identify.

feiranl (approved):
Besides the comment above, all others looks good to me. Well done!

model/Human-GEM.yml
Devlin-Moyer: In my opinion, only one version should be left unblocked by default. Until this thread, it didn't even occur to me that Complex IV effectively has twice as much flux capacity as the rest of the ETC reactions by default, since the other ETC reactions don't (as far as I know) have comparable ROS-producing duplicate versions. Nothing about MAR06914 suggests that it has an ROS-producing duplicate in MAR13081, and I think it'd it'd be much harder to figure out why it seems like there's more flux than you bargained for going through Complex IV (e.g. you only changed the bound on MAR06914 and did not know MAR13081 existed) than to figure out why Complex IV isn't producing ROS (which you can figure out easily by just looking at the reaction string for MAR06914 and seeing no ROS production)

JonathanRob: I agree with @Devlin-Moyer, this reaction can easily be missed, and may cause confusion if one is not aware of the two pathways. I can't think of a scenario where deactivating it would cause any issues, but I can think of a few annoyances that could be caused by keeping both reactions open. So I would lean toward deactivating it.

Devlin-Moyer: Also, at the moment, nobody seems to know where the stoichiometric coefficients in MAR13081 came from, but the stoichiometric coefficients in MAR06914 clearly come from [the associated reference](https://www.sciencedirect.com/science/article/pii/0304417386900066?via%3Dihub)

feiranl: Ok, in that case, could we change the name or the subsystem to indicate the reason why we block this at least point to ROS or similar commet. Otherwise, in one or two years, we forget why we block this rxn.

JonathanRob: @feiranl the reason the model is maintained on GitHub is precisely so that in 2 years (or more) we can search this repository to find out why changes were made. I would not recommend changing the subsystem just for tracking reasons, but if there is another subsystem which is actually deemed more appropriate, then of course that change can be considered. It's also fine to update the name of the reaction to be more descriptive, if necessary.

Issue #527, opened 2023-04-05
(d)NTP <-> (d)NMP + PPi Reactions Should Be Irreversible


Devlin-Moyer:
Technically, the reaction ATP + H<sub>2</sub>O <-> AMP + PP<sub>i</sub> is almost perfectly reversible at physiological temperatures and pHs, but all cells express pyrophosphatases that rapidly hydrolyze all pyrophosphate into free inorganic phosphates (PP<sub>i</sub> + H<sub>2</sub>O -> 2 P<sub>i</sub>; a very irreversible reaction), specifically so that all of these pyrophosphate-producing reactions are driven in the pyrophosphate-producing direction and wind up being effectively irreversible ([source](https://doi.org/10.1007/978-3-642-58444-2_7)).
In Human1, most reactions that have a (d)NTP on one side and pyrophosphate on the other are reversible, and while this is technically not wrong, it allows FBA to predict very large ATP-producing fluxes from unrealistic sources, because FBA is under no obligation to ensure that most of the pyrophosphate produced from (d)NTPs winds up going through the pyrophosphatase reactions (MAR03975, MAR03977, MAR03979, MAR06912, MAR07794) if it can also run all these ATP <-> AMP + PP<sub>i</sub> reactions in the ATP-producing direction.
There are quite a few reactions like this, so I used this bit of code to find all of them and set either the upper or lower bound to zero:

```python
for r in model.reactions:
    # skip the pyrophosphate exchange and pyrophosphatase reactions
    pi_in_rxn = any(m.id.startswith('MAM02751') for m in r.metabolites)
    # also skip reactions that are already irreversible
    if (r not in model.boundary) and not pi_in_rxn and r.reversibility:
        # figure out if pyrophosphate is a product or reactant (or both)
        ppi_prod = any(m.id.startswith('MAM02759') for m in r.products)
        ppi_reac = any(m.id.startswith('MAM02759') for m in r.reactants)
        if ppi_prod and not ppi_reac:
            r.lower_bound = 0
        elif ppi_reac and not ppi_prod:
            r.upper_bound = 0
```

<details>
  <summary>Upper Bounds Should Be Zero</summary>
  MAR00156
  MAR00165
  MAR00168
  MAR00171
  MAR00174
  MAR00177
  MAR00180
  MAR00184
  MAR00188
  MAR00192
  MAR00196
  MAR00209
  MAR00213
  MAR00217
  MAR00226
  MAR00233
  MAR00237
[...]

PR #557, opened 2023-05-08
Fix: Make all non-transport/exchange pyrophosphate reactions irreversible

Devlin-Moyer:
Makes all reversible non-transport/exchange reactions that involve pyrophosphate (MAM02759e/c/m/x/n/r) irreversible in the direction of pyrophosphate production, as discussed in #527. In cases where pyrophosphate is currently a “reactant,” products and reactants are swapped to ensure no reactions are constrained to only have fluxes <= 0.

### DIFF
--- a/.github/workflows/check-metabolictasks.yml
+++ b/.github/workflows/check-metabolictasks.yml
@@ -1,0 +1,19 @@
+name: Test metabolic tasks
+
+on: [push, pull_request]
+
+jobs:
+  check-metabolictasks:
+    runs-on: self-hosted
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        task-type: [essential, verification]
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Check ${{ matrix.task-type }} metabolic tasks
+      run: |
+        /usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); testMetabolicTasks('${{ matrix.task-type }}'); exit;"
--- a/.github/workflows/yaml-conversion.yml
+++ b/.github/workflows/yaml-conversion.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
 
     - name: Checkout
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: Run conversion script
       run: |
--- a/code/test/sanityCheck.py
+++ b/code/test/sanityCheck.py
@@ -28,7 +28,7 @@ def load_yml(yml_file):
     modelMets  = list(map(lambda element : element.id, model.metabolites))
     modelGenes = list(map(lambda element : element.id, model.genes))
 
-    return modelRxns, modelMets, modelGenes
+    return modelRxns, modelMets, modelGenes, model
 
 
 def checkRxnAnnotation(rxns):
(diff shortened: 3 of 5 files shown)


Devlin-Moyer:
Oh dang; I used Cobrapy's built-in function for saving models in YAML format, but it appears to have not used the format your automated format checker was expecting

haowang-bioinfo:
looks good

I'm afraid the effect of these changes needs to be checked with verification and essentiality tests that is under construction, will come back to as soon as possible

haowang-bioinfo:
> Oh dang; I used Cobrapy's built-in function for saving models in YAML format, but it appears to have not used the format your automated format checker was expecting

actually is considering to fix the compatibility to cobrapy, any ideas in improving this?

Devlin-Moyer:
Cobrapy uses [ruamel.yaml](https://yaml.readthedocs.io/en/latest/) to [generate their YAML files](https://cobrapy.readthedocs.io/en/latest/_modules/cobra/io/yaml.html#save_yaml_model). I'm not certain if these are all of the differences, but:
* Cobrapy's YAML files don't have the header section that yours does
* Cobrapy's YAML files don't quote all character/text/non-numeric fields by default, while yours do (without the quotes, your automated validation bot kept interpreting the "formula" entires for nitrous oxide (NO) metabolites as if they were supposed to be True/False and telling me it couldn't interpret "NO" as either "True" nor "False")
* Cobrapy's YAML files use two spaces for each level of indentation, but yours have 2 spaces for the first level, 4 spaces for the second, 2 for the third, etc.
* Cobrapy's YAML files have a separate "id" section (for the model id) at the "root" level (i.e. the same level as the metabolites, reactions, genes, and compartments)
* Cobrapy's YAML files seem to wrap/split lines that are over 80 characters across multiple lines, which only seemed to affect metabolites with long names and reactions with long gene_reaction_rules (also your automated validation bot never flagged that as an issue)

I suspect that the biggest issues are the contents of the header and the quotes around metabolite formulas; I think fixing the header problem would require changes on Cobrapy's end, because, as far as I can tell, most of the information you have in the header of your YAML file isn't in a Cobrapy YAML file at all. The other things seem like things you could probably tweak with your YAML validation thing to accommodate, but I'm not entirely sure how it works

mihai-sysbio:
While I realise it is more cumbersome, SBML is a more robust file format for using the model across different tools/platforms.

haowang-bioinfo:
> it may be worth noting that Cobrapy had no problem reading in Human-GEM.yml, as far as I could tell

yes, the reading was fixed early on; now probably is time to resolve writing side

> Cobrapy's YAML files don't have the header section that yours does
Cobrapy's YAML files don't quote all character/text/non-numeric fields by default, while yours do (without the quotes, your automated validation bot kept interpreting the "formula" entires for nitrous oxide (NO) metabolites as if they were supposed to be True/False and telling me it couldn't interpret "NO" as either "True" nor "False")
Cobrapy's YAML files use two spaces for each level of indentation, but yours have 2 spaces for the first level, 4 spaces for the second, 2 for the third, etc.
Cobrapy's YAML files have a separate "id" section (for the model id) at the "root" level (i.e. the same level as the metabolites, reactions, genes, and compartments)
Cobrapy's YAML files seem to wrap/split lines that are over 80 characters across multiple lines, which only seemed to affect metabolites with long names and reactions with long gene_reaction_rules (also your automated validation bot never flagged that as an issue)

thanks for listing the details

> I suspect that the biggest issues are the contents of the header and the quotes around metabolite formulas; I think fixing the header problem would require changes on Cobrapy's end, because, as far as I can tell, most of the information you have in the header of your YAML file isn't in a Cobrapy YAML file at all. The other things seem like things you could probably tweak with your YAML validation thing to accommodate, but I'm not entirely sure how it works.

the quote issue is technically solvable, while header section actually is the tricky part

haowang-bioinfo:
> I realise it is more cumbersome, SBML is a more robust file format for using the model across different tools/platforms.

yes SBML is a mess

mihai-sysbio:
> would require changes on Cobrapy's end

This might be the better solution in more than one case.

haowang-bioinfo:
> I'm afraid the effect of these changes needs to be checked with verification and essentiality tests that are under construction, will come back as soon as possible

@Devlin-Moyer would you be able to merge the latest `develop` branch into this PR, or start a new PR from the latest `develop` to add the commits (then close this PR). The idea is to make sure that these changes won't break the model

haowang-bioinfo (approved):
very nice - significant progress in improving thermodynamic constraints!

PR #564, opened 2023-05-12
feat: add test for metabolic tasks

haowang-bioinfo:
This PR adds GH action check for both essential and basic verification tasks.

Note: ~~the GH workflow appears to be recalcitrant to accept multiple variables provided in the action yaml file, no sure if is a bug from GH action or Yaml syntax~~ The `run` command in Matlab actually can only deal with script, not for function with parameters.

### DIFF
--- a/.github/workflows/check-metabolictasks.yml
+++ b/.github/workflows/check-metabolictasks.yml
@@ -6,14 +6,14 @@ jobs:
   check-metabolictasks:
     runs-on: self-hosted
     timeout-minutes: 60
-    env:
-      matrix:
-        param: [essential, verification]
     steps:
 
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Check essential tasks
+    - name: Check metabolic tasks
+      env:
+        matrix:
+          param: [essential, verification]
       run: |
         /usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); run('code/test/testMetabolicTasks.m', '${{ env.matrix.param }}'); exit;"
--- a/.github/workflows/check-metabolictasks.yml
+++ b/.github/workflows/check-metabolictasks.yml
@@ -2,6 +2,10 @@ name: Test metabolic tasks
 
 on: [push, pull_request]
 
+env:
+  matrix:
+    param: [essential, verification]
+
 jobs:
   check-metabolictasks:
     runs-on: self-hosted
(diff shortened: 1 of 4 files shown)


mihai-sysbio:
For improved maintainability, I would recommend adding this check as a new job next to the [`yaml-conversion` job](https://github.com/SysBioChalmers/Human-GEM/blob/main/.github/workflows/yaml-conversion.yml#L6). The filename and `name` entry in the file should also be changed. [Jobs run in parallel by default](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow).

.github/workflows/check-metabolictasks.yml
```diff
@@ -0,0 +1,27 @@
...
+      env:
+        matrix:
+          param: 'essential'
+
+    - name: Check verification tasks
+      run: |
+        /usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); run('code/test/testMetabolicTasks.m', '${{ matrix.param }}'); exit;"
+      env:
+        matrix:
+          param: 'verification'
```
mihai-sysbio, line 27: I guess the purpose here was to [use a matrix like in the docs](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#using-a-matrix-strategy):
```suggestion
          param: [essential, verification]
```

mihai-sysbio:
Really nice work @haowang-bioinfo 🚀

mihai-sysbio:
@haowang-bioinfo have a look at this action run and see if you like how the matrix setup works https://github.com/SysBioChalmers/Human-GEM/actions/runs/4974385471 .

By the way, I would still prefer to merge the workflow files as mentioned above.

There still seems to be a problem with the command though (copied from the action run), could you try to see if it runs on your machine?
```
/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "addpath(genpath('.')); run('code/test/testMetabolicTasks.m', 'verification'); exit;"
```

haowang-bioinfo:
> There still seems to be a problem with the command though (copied from the action run), could you try to see if it runs on your machine?

thanks for checking, will check this locally

.github/workflows/check-metabolictasks.yml
```diff
@@ -0,0 +1,19 @@
...
+    timeout-minutes: 60
+    strategy:
+      matrix:
+        task-type: [essential, verification]
+    steps:
+
+    - name: Checkout
+      uses: actions/checkout@v2
+
+    - name: Check metabolic tasks
```
mihai-sysbio, line 17: The task type can also be used in the step name:
```suggestion
    - name: Check ${{ matrix.task-type }} metabolic tasks
```

mihai-sysbio (approved):
Very nice to have this in the repo!

PR #582, opened 2023-05-25
feat: add the duplicate rxn check task

feiranl:
This PR adds GH action check for duplicate reactions by `unique` and `count` the reaction equations with metabolite IDs.

Now this PR is ready to be merged, after resolving duplicates by:
- merge MAR20012 into MAR06646
- merge MAR20089 into MAR04545

**The task for YAML validation fails because there are currently duplicate reactions in the model. I have manually tested the function on Yeast-GEM, and it was successful. 
Please wait until all duplicate reactions are removed in the devel branch and merged back to this branch. Once the model passes the check, we can merge this pull request into the devel branch.**


haowang-bioinfo:
convert to `draft` to avoid merging

feiranl:
The `duplicateRxnCheck` has not been merged into the develop branch yet. We are currently waiting for the import and export cycle to be fixed. Once the duplicate reactions (#630) are removed, the check implemented in this branch will be successful, and then we can proceed with the merge.

haowang-bioinfo:
now there are duplicate reactions found in #709, but escaped from basic memote check #672

might be good to add this version of check as test, and it's okay have both

haowang-bioinfo:
The failed Yaml-validation action reports:
```
The following 14 reactions are duplicates, please check: MAR04521;MAR06981;MAR04689;MAR06631;MAR06646;MAR03984;MAR00067;MAR20002;MAR20012;MAR20014;MAR20015;MAR20020;MAR20052;MAR20066
```

mihai-sysbio:
> it's okay have both

These two should ideally have identical behaviour, so keeping one would be best. It would mean less maintenance, and it would also make it clearer to have 1 single failing task instead of 2, given the same error.

haowang-bioinfo:
>> it's okay have both
>>
> These two should ideally have identical behaviour, so keeping one would be best. It would mean less maintenance, and it would also make it clearer to have 1 single failing task instead of 2, given the same error.

yes, one check would be ideal for less maintenance, if the two behave identically.  
now they behave differently and appear to complement each other, so it's okay to keep both.

haowang-bioinfo (approved):

mihai-sysbio:
Could we hold off the merging for a little bit? I'm running the Memote check locally to see if indeed it is misbehaving, or if it is just a reporting issue.

haowang-bioinfo:
> Could we hold off the merging for a little bit? 

sure, let's hold

> I'm running the Memote check locally to see if indeed it is misbehaving, or if it is just a reporting issue.

Memote check and this PR are not conflicting, they may supplement each other

mihai-sysbio:
Having found the source the issue, I agree these are not conflicting.

In Memote, duplication of reactions is not based on identifiers but on inchis of the metabolites ([source](https://github.com/opencobra/memote/blob/3687f3734c5a864b399e99c7aa26e180e8830ff6/src/memote/support/basic.py#L470)). So I believe that if the yml file would be populated with inchis, the Memote check would be more thorough.

Please proceed how you see fit.

code/test/sanityCheck.py
```diff
@@ -106,11 +107,28 @@ def checkUnusedEntities(model, entity_type):
     assert len(unused_entities) == 0, f"Found unused {entity_type}: {unused_entities}"
 
 
+def checkDupRxn(model):
+    """
+    Check for duplicate reactions in the model
+    """
+
+    reaction_equations = [rxn.build_reaction_string(use_metabolite_names=False) for rxn in model.reactions]
```
mihai-sysbio, line 115: I suspect that this would only catch duplicated reactions where the metabolites have the exact same order?

haowang-bioinfo: I assume metabolites are often sorted, so that duplicated reactions can be detected

haowang-bioinfo: the output might be further optimized by showing duplicate pairs in groups, instead of a list of rxn ids

Issue #588, opened 2023-05-27
GPRs for MAR09985 and MAR11307


Devlin-Moyer:
### Background:
Both reactions currently have no genes associated with them:
`MAR09985: 5-aminolevulinate [c] + 2 H+ [c] ⇔ 5-aminolevulinate [e] + 2 H+ [e]`
`MAR11307: 5-aminolevulinate [c] + H+ [c] ⇔ 5-aminolevulinate [e] + H+ [e]`
According to Uniprot, [SLC15A1](https://www.uniprot.org/uniprotkb/P46059/entry) (`ENSG00000088386`) can catalyze either of these reactions, and [SLC36A1](https://www.uniprot.org/uniprotkb/Q7Z2H8/entry) (`ENSG00000123643`) can catalyze `MAR11307` but not `MAR09985`. Both of these genes are already in Human-GEM and associated with many other similar reactions.

### Proposed Changes:
- [ ] Add `ENSG00000088386` to the currently empty GPR of `MAR09985`
- [ ] Add `ENSG00000088386 or ENSG00000123643` to the currently empty GPR of `MAR11307`

haowang-bioinfo:
based on catalytic activity of [SLC15A1](https://www.uniprot.org/uniprotkb/P46059/entry) and [SLC36A1](https://www.uniprot.org/uniprotkb/Q7Z2H8/entry), direct evidence supporting 5-aminolevulinate as substrate is lacking

Devlin-Moyer:
Oh whoops I meant to cite [this paper](https://jpet.aspetjournals.org/content/332/1/220.short?casa_token=ew1LMkK3PL8AAAAA:aOt6JDSQHiO1WlFZKmEFWSu5Ll3pBZibrtU4N5Nvd_O9BXxqo908d8QY70XaB09ZFZ5SZ7YLTkDG) and not just Uniprot for this

haowang-bioinfo:
> Oh whoops I meant to cite [this paper](https://jpet.aspetjournals.org/content/332/1/220.short?casa_token=ew1LMkK3PL8AAAAA:aOt6JDSQHiO1WlFZKmEFWSu5Ll3pBZibrtU4N5Nvd_O9BXxqo908d8QY70XaB09ZFZ5SZ7YLTkDG) and not just Uniprot for this

yes, this paper reports that SLC15A1 and SLC36A1 contribute to proton-coupled transport of 5-aminolevulinate in the intestine.

[...]

PR #623, opened 2023-06-08
fix: add GPRs and references for MAR09985 and MAR11307

Devlin-Moyer:
As discussed in #588:
* Adds `ENSG00000088386` to the currently empty GPR of `MAR09985`
* Adds `ENSG00000088386 or ENSG00000123643` to the currently empty GPR of `MAR11307`

Also, while this was not proposed in #588, while I was editing those reactions, I figured I'd add `PMID:19789362` to the references for both reactions, and I noticed that both reactions already had references, namely [PMID:11752223](https://pubmed.ncbi.nlm.nih.gov/11752223/) for `MAR09985` and [PMID:21749716](https://europepmc.org/article/PMC/3158119) for `MAR11307`. However, as far as I can tell, neither of those papers actually provide evidence of transport of 5-aminolevulinate (`MAM01074e/c`), so I replaced both of those papers with `PMID:19789362`, since, as discussed in #588, that paper does actually provide evidence that supports transport of 5-aminolevulinate, specifically via SLC15A1 (`ENSG00000088386`) and SLC36A1 (`ENSG00000123643`)


haowang-bioinfo (approved):
really appreciate the efforts of updating the references

Issue #562, opened 2023-05-10
Pairs of irreversible transport reactions that should probably be merged into single reversible reactions


Devlin-Moyer:
I used this bit of Python code to systematically identify pairs of irreversible transport reactions oriented in opposite directions, and I think most of them should be combined into single reversible reactions. I was only trying to get transport reactions, but this code also happens to have found a few pairs of irreversible NADH- or NADPH-dependent redox reactions that also seem like they should be combined into single reversible reactions (those reactions are now mentioned in separate issues: #578, #579, and #580).

<details>
  <summary>Python Code</summary>

```python
import cobra
import itertools as it

# these functions & class take a GPR and return a frozenset of
# frozensets, where each frozenset is a list of genes that can
# independently catalyze the reaction, e.g.
# solve('A') -> {{'A'}}
# solve('A or B') -> {{'A'}, {'B'}}
# solve('A and B') -> {{'A', 'B'}}
# solve('A and (B or C)') -> {{'A', 'B'}, {'A', 'C'}}
# solve('(A and B) or (C and D)') -> {{'A', 'B'}, {'C', 'D'}}
def frozenproduct(first, second):
   for e1 in first:
      for e2 in second:
         yield e1.union(e2)

class BoolBlock:
   ops = ['and', 'or']

   def __init__(self):
      self.op = None
      self.possible = set()

   def settype(self, token):
      assert token in BoolBlock.ops
      if self.op is None:
         self.op = token
      else:
         assert self.op == token

   def add(self, new):
      if self.op == 'and':
         self.possible = set(frozenproduct(self.possible, new))
      else:
         self.possible.update(new)

def tokenize(string):
   buffer = ''
   for c in string:
      if c in ' ()':
         if buffer.strip():
            yield buffer.strip()
         if c.strip():
            yield c
         buffer = ''
      else:
         buffer += c

   if buffer.strip():
      yield buffer.strip()

def recurse(token_stream):
   current = BoolBlock()
   for token in token_stream:
      if token == '(':
         current.add(recurse(token_stream).possible)
      elif token == ')':
         return current
      elif token in BoolBlock.ops:
         current.settype(token)
      else:
         gene = {frozenset({token,}), }
         current.add(gene)
   return current

def solve(string):
   return recurse(tokenize(string)).possible

model = cobra.io.load_yaml_model('Human-GEM.yml')

bad_pairs = set()
# loop over metabolites
for ref_met in list(model.metabolites):
[...]

PR #625, opened 2023-06-08
Merge Pairs Of Irreversible Transport Reactions

Devlin-Moyer:
For each pair of reactions mentioned in #562:
* Makes the reaction with the lower/smaller ID number reversible
* Copies any E.C. codes and references from the reaction with the lower/smaller ID number to the reaction with the higher/larger ID number
* Removes the reaction with the higher/larger ID number


feiranl:
Nice! but it would be great if you can also merge the cross reference annotations and rxnNotes in the model yaml file and `reactions.tsv` file for the further usage. I am working on the tissue specific tasks. Since recently we removed many rxns, it is difficult to find the `kept rxn` from the old documented rxn ID of the duplicate pair without manually search in the issue.

Devlin-Moyer:
I will start working on that

Devlin-Moyer:
FYI none of the removed reactions had any rxnNotes, but several had EC codes that the kept reactions didn't have, so I copied over both the references and the EC codes

haowang-bioinfo:
@Devlin-Moyer good work

just a little concern about the merging: what if it turns out that some import and export reactions of the same metabolites are later associated with different transporters?

Devlin-Moyer:
At the moment, we have no particular reason to believe that's the case (as far as I know -- correct me if I'm wrong), and the presence of these paired irreversible opposite reactions creates a bunch of loops of unbounded fluxes in the model, so it seems reasonable to merge them all now and only separate them back out again when we have a specific reason to intentionally introduce a loop into the model

haowang-bioinfo:
> the presence of these paired irreversible opposite reactions creates a bunch of loops of unbounded fluxes in the model

could you please list these cases

Devlin-Moyer:
Every single pair of reactions in #562 forms a loop that can sustain arbitrarily large flux, since, in each pair of reactions, one has exactly the opposite stoichiometry as the other. I first came across these reactions by getting representative samples of solutions to Human-GEM-derived models and looking for reactions that had average possible fluxes that were suspiciously large (>100 times larger than any bound I had set on an exchange reaction)

Devlin-Moyer:
Since there are only two reactions involved in each, they're very self-contained, but I've been spending a lot of time sampling from solution spaces of models, and each reaction with an unbounded flux is a dimension of the solution space that extends infinitely without ever hitting a bound, which makes the whole process slightly more annoying than it needs to be

haowang-bioinfo:
>> what if it turns out that some import and export reactions of the same metabolites are later associated with different transporters?
>
> At the moment, we have no particular reason to believe that's the case

ture, there is no evidence

> Every single pair of reactions in #562 forms a loop that can sustain arbitrarily large flux, since, in each pair of reactions, both have identical metabolites, and one has exactly the opposite stoichiometry as the other.

yes indeed

> each reaction with an unbounded flux is a dimension of the solution space that extends infinitely without ever hitting a bound, which makes the whole process slightly more annoying than it needs to be

make sense to me

feiranl:
I am ok with the change. As for the unbounded flux, it may be solved by using pFBA @Devlin-Moyer

haowang-bioinfo (approved):

haowang-bioinfo (approved):

haowang-bioinfo:
look good!

Issue #628, opened 2023-06-12
bug identified in development branch


feiranl:
import and export cycle gives an error:
```
model = importYaml('Human-GEM.yml');
>> exportYaml(model, 'Human-GEM.yml');
Index exceeds the number of array elements. Index must not exceed 13087.

Error in exportYaml>writeField (line 190)
            list = field{pos};  %subSystems already comes in a cell array

Error in exportYaml (line 86)
    writeField(model, fid, 'subSystems',           'txt', i, '  - subsystem')
```

This is due to the subsystems length is different with other rxn related field.
<img width="358" alt="Screenshot 2023-06-12 at 10 51 09 AM" src="https://github.com/SysBioChalmers/Human-GEM/assets/32157802/64ad23a3-84c6-4113-a224-f7c25dea4535">

mihai-sysbio:
Oh that's surprising to see. There is a GH workflow that runs the [yaml conversion check](https://github.com/SysBioChalmers/Human-GEM/blob/main/code/test/testYamlConversion.m). It looks like when the test fails the workflow still passes - have a look at the step _Run conversion script_ for [this workflow ron](https://github.com/SysBioChalmers/Human-GEM/actions/runs/5237189388/jobs/9455204803) on the merged PR. Looking back through the results, the problem might have been introduced in https://github.com/SysBioChalmers/Human-GEM/pull/606.

mihai-sysbio:
My suggestion is that the fix should also include a change in the test such that when it fails the action also fails. After this, the fix of the yaml file should be committed, and we should see the workflow pass.

feiranl:
it errors in the check, but didnot fail in the [action](https://github.com/SysBioChalmers/Human-GEM/pull/606/checks):


                                           

 
{Index exceeds the number of array elements (13087).

Error in exportYaml>writeField (line 190)
            list = field{pos};  %subSystems already comes in a cell array

Error in exportYaml (line 86)
    writeField(model, fid, 'subSystems',           'txt', i, '  - subsystem')

Error in testYamlConversion (line 25)
exportYaml(model,'testYamlConversion.yml');

Error in run (line 91)
evalin('caller', strcat(script, ';'));
} 

@haowang-bioinfo maybe you can fix this?

haowang-bioinfo:
> @haowang-bioinfo maybe you can fix this?

yes, working on this

haowang-bioinfo:
[...]

PR #629, opened 2023-06-12
fix-add a error report in the import/export cycle

feiranl:
## Main improvements in this PR:
This PR is to fix only the bug found in Yaml file, as reported in #628. In the meantime, another PR #635 is to fix error of GH workflow test

### DIFF
--- a/code/io/importYaml.m
+++ b/code/io/importYaml.m
@@ -226,7 +226,7 @@ for i=1:numel(line_key)
                 
             otherwise
                 if readSubsystems
-                    subSystems(end+1,1) = {regexprep(tline_value, '^ *- (.+)$','$1')};
+                    subSystems = {regexprep(tline_value, '^ *- (.+)$','$1')};
 
                 % resolve the equation
                 elseif readEquation
--- a/code/io/importYaml.m
+++ b/code/io/importYaml.m
@@ -226,7 +226,7 @@ for i=1:numel(line_key)
                 
             otherwise
                 if readSubsystems
-                    subSystems = {regexprep(tline_value, '^ *- (.+)$','$1')};
+                    subSystems(end+1,1) = {regexprep(tline_value, '^ *- (.+)$','$1')};
 
                 % resolve the equation
                 elseif readEquation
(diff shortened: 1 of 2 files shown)


feiranl:
@haowang-bioinfo Could you help to fix this? I add an error catch, but the task still succeed.

haowang-bioinfo:
> I propose a error message to fix the bug.

the issue presumably is due to using previously cached model by the GitHub action cloud server, and a probable solution is to "force" GH action to use freshly loaded model every time. What do you think @mihai-sysbio?

mihai-sysbio:
I'm confused - the test is now successful. Is that right?

My idea was to start by fixing the test first, so that we can confirm that the test will trigger a failure of the GH workflow. Only after that the yaml problem would be fixed, and the re-run of the workflow would then pass.

mihai-sysbio:
I renamed the branch to `fix/importExportCycle`, which then triggered a closure of the PR and deletion of the branch.

feiranl:
It is not ready for closing. There is another bug I identified in importYML. 

the import and export cycle changes the yml file.
MicrosoftTeams-image

mihai-sysbio:
> It is not ready for closing

Absolutely! It was not the intention to close the PR, sorry about that. Normally branch renaming is seamless, but apparently when a PR is already open, GitHub will close it. Let's continue in #635.

haowang-bioinfo:
> There is another bug I identified in importYaml.
the import and export cycle changes the yml file.

yes indeed, working on this now

haowang-bioinfo:
>> There is another bug I identified in importYaml.
the import and export cycle changes the yml file.
>
> yes indeed, working on this now

this bug is located from Yaml file line 307295: `- metabolties: !!omap`, where there is typo that supposed to be `metabolites`. This particular bug was introduced in #606

haowang-bioinfo:
this PR is reopened for fixing yaml import/export cycle only, while the issue of GH workflow test will be fixed in #635

feiranl:
LGTM! The import and export cycle works fine now

haowang-bioinfo (approved):

PR #636, opened 2023-06-13
fix: simplifyGrRules: The old code didn't run, something has probably changed.

johan-gson:
Fixed a bug in simplifyGrRules, it couldn't be run. It is not as critical as the title implies, I thought there was something worse going on :)


mihai-sysbio:
This PR is a duplicate of #494. Perhaps you could review that and then it can be merged?

johan-gson:
@mihai-sysbio Could you delete the branch associated with this, it is no longer needed? I don't think I have the right to do that.

PR #613, opened 2023-06-07
Fix FAD/Ubiquinone Duplicates

Devlin-Moyer:
#### Main improvements in this PR:

As discussed in #607:
- Removes `MAR02035` for being a duplicate of `MAR03121`
- Removes `MAR03211` for being a duplicate of `MAR03212`
- Removes `MAR03751` for being a duplicate of `MAR03752`
- Removes `MAR03783` for being a duplicate of `MAR03784`
- Removes `MAR04242` for being a duplicate of `MAR04243`
- Removes `MAR03769` for being a duplicate of `MAR03770`
- Removes `MAR03838` for being a duplicate of `MAR08611`
- Removes `MAR02366` for being a duplicate of `MAR03136`
- Removes `MAR02367` for being a duplicate of `MAR03135`
- Removes `MAR02370` for being a duplicate of `MAR03128`
- Removes `MAR02372` for being a duplicate of `MAR03156`
- Removes `MAR02369` for being a duplicate of `MAR03149`
- Removes `MAR02373` for being a duplicate of `MAR03142`
- Removes `MAR00483` for being a duplicate of `MAR00482`/`MAR00449`/`MAR01169`
- Removes `MAR08743` for being a duplicate of `MAR04652`

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [X] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

Issue #607, opened 2023-06-06
FAD/Ubiquinone Duplicate Reactions


Devlin-Moyer:
### The Problem:

These pairs of reactions all represent reactions that are catalyzed by enzymes that use FAD as a cofactor, and the electrons harvested from their substrates eventually make their way from FAD to ubiquinone, but Human-GEM currently has one version that involves FAD and another that directly reduces ubiquinone with no explicit acknowledgement of the role of FAD:

substrates | ubiquinone-dependent ID | FAD-dependent ID
---|---|---
palmitoyl-CoA, (2E)-hexadecenoyl-CoA | MAR02035 | MAR03121
propanoyl-CoA, acrylyl-CoA | MAR03211 | MAR03212
isobutyryl-CoA, methacrylyl-CoA | MAR03751 | MAR03752
2-methylbutyryl-CoA, tiglyl-CoA | MAR03783 | MAR03874
glutaryl-CoA, crotonoyl-CoA | MAR04242 | MAR04243
3-methylcrotonoyl-CoA, isovaleryl-CoA | MAR03769 | MAR03770
butyryl-CoA, crotonoyl-CoA | MAR02366 | MAR03136
lauroyl-CoA, (2E)-dodecenoyl-CoA | MAR02367 | MAR03135
myristoyl-CoA, (2E)-tetradecenoyl-CoA | MAR02370 | MAR03128
hexanoyl-CoA, (2E)-hexenoyl-CoA | MAR02372 | MAR03156
decanoyl-CoA, (2E)-decenoyl-CoA | MAR02373 | MAR03142
1-pyrroline-5-carboxylate, proline | MAR03838 | MAR08611
fumarate, succinate | MAR04652 | MAR08743
DHAP, sn-glycerol-3-phosphate | MAR00483 | MAR00482, MAR00449, MAR01169 (see #599)

Aside from the last few reactions, if you look at any of the genes associated with any of these reactions, the overwhelming majority of other reactions associated with thse genes only have FAD-dependent versions (e.g. out of the 90 reactions ACADVL is associated with, `MAR02035` is the only one that uses ubiquinone instead of FAD), so for consistency, it would make sense to just drop the ubiquinone-dependent member of each of these pairs.

Speaking of the genes associated with these reaction, some of them (e.g. ACOX1 and ACOX3) are peroxisomal and not mitochondrial, but they're also associated with a bunch of other mitochondrial reactions, so I'm making a separate issue about that.

[...]

PR #663, opened 2023-06-21
Fix duplicate metabolites identified by HMDB IDs

JHL-452b:
#### Main improvements in this PR:

As discussed in https://github.com/SysBioChalmers/Human-GEM/issues/617
- remove mets `MAM03318e` for being duplicates
- remove rxns `MAR02276`, `MAR02277`, `MAR02384`, `MAR02390`, `MAR02392`, `MAR02394`, `MAR04144 `and `MAR10447` for being affected by duplicate mets
- change HMDB ID annotations for `MAM02679e`, `MAM01090c `and `MAM01090m`
- replace mets

|old ID|new ID| 
|-----------------|-----------------|
|MAM03318c| MAM01435c|
|MAM03318r| MAM01435r|
|MAM03318m| MAM01435m|
|MAM03325c| MAM00684c|

**I hereby confirm that I have:**

- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [X] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

PR #646, opened 2023-06-18
Fix duplicate metabolites identified by BiGG IDs

JHL-452b:
#### Main improvements in this PR:

As discussed in https://github.com/SysBioChalmers/Human-GEM/issues/610

-  remove mets `MAM03226m`, `MAM03884c`, `MAM03884e`, `MAM03554c` for being duplicates
-  remove rxns `MAR00042`,`MAR01617`, `MAR00740`, `MAR00659`, `MAR05323`, `MAR01709 `for being affected by duplicate mets
-  change BiGG ID annotations for `MAM01746c`, `MAM00078x`, `MAM02328c`, `MAM02328g`, `MAM02328l`, `MAM02328e`, `MAM02328r`, `MAM02008c`, `MAM02008g`, `MAM02008l`, `MAM02008e `
- replace mets

 |old ID|new ID| 
|-----------------|-----------------|
|MAM03884x| MAM02746x|

**I hereby confirm that I have:**

- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [X] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

PR #666, opened 2023-06-23
feat: log duplicated ids to the column of rxnRetired

haowang-bioinfo:
This PR improves documentation of deprecated reactions in #613 by adding their ids to corresponding rows in `reactions.tsv`, as a tentative implementation according to the discussion in #615

Note: all the pairs listed in #607 have been recorded except `MAR00483 vs. MAR00482/MAR00449/MAR01169`, because I don't know how to document this irregular pair


model/reactions.tsv
```diff
@@ -819,7 +819,7 @@ rxns	rxnKEGGID	rxnBiGGID	rxnEHMNID	rxnHepatoNET1ID	rxnREACTOMEID	rxnRecon3DID	rx
 "MAR04232"	""	""	"RE0549C"	""	""	"RE0549C"	"MNXR103449"	"HMR_4232"	"RCR10536"	""	0	""	""	"HMR_4232"
 "MAR04233"	"R03889"	""	""	"r0645"	""	"r0645"	"MNXR95796"	"HMR_4233"	"RCR11396"	""	0	""	"RHEA:14469"	"HMR_4233"
 "MAR04235"	"R01938"	"AMCOXO"	""	"r0449"	""	"AMCOXO"	"MNXR95807"	"HMR_4235"	"RCR11397"	""	0	""	""	"HMR_4235"
-"MAR04243"	"R02487"	"GLUTCOADHm"	"R02487M"	"r0541"	""	"GLUTCOADHm;r0541"	"MNXR100293"	"HMR_4243;HMR_4242"	"RCR11398;RCR14356"	""	0	""	"RHEA:30847"	"HMR_4243;HMR_4242"
+"MAR04243"	"R02487"	"GLUTCOADHm"	"R02487M"	"r0541"	""	"GLUTCOADHm;r0541"	"MNXR100293"	"HMR_4243;HMR_4242;MAR04242"	"RCR11398;RCR14356"	""	0	""	"RHEA:30847"	"HMR_4243;HMR_4242"
```
feiranl, line 822: I am not sure about the reason this one appears not in the last column

feiranl (approved):
LGTM! Except the pair of MAR03784 and MAR03783 is not documented. We also performed this kind of documentation in the #646 and should performed in every PR with removing duplicates both for Rxns and Mets

model/reactions.tsv
haowang-bioinfo: this is actually my mistake, now is corrected 9ce2c5c8e2a77dacd46850d788b4e661be399005

haowang-bioinfo:
> Except the pair of MAR03784 and MAR03783 is not documented.

thank you for the careful inspection that really helps in avoiding mistakes!

> We also performed this kind of documentation in the https://github.com/SysBioChalmers/Human-GEM/pull/646 and should performed in every PR with removing duplicates both for Rxns and Mets

yes, we probably follow this practice for all pending PRs and onward

mihai-sysbio:
From my side, this requires a bit more thinking. Let's take `MAR03784` as an example:
- what is the meaning that for a given reaction the same entries `HMR_3784;HMR_3783` exist in two columns `rxnHMR2ID` and `rxnRetired`?
- even though we now know if `MAR03783` has been retired, there is no reference as to who/what/when, so in essence the core dilemma persists
- the TSV files in `model/` are used for id mapping, but this new column doesn't quite have the same purpose
- for accuracy, shouldn't one look back in past PRs to populate this new column?

haowang-bioinfo:
> what is the meaning that for a given reaction the same entries HMR_3784;HMR_3783 exist in two columns rxnHMR2ID and rxnRetired?

no immediate answer, but this can definitely be figured out given the radically transparent history of data/code in Human-GEM (an investigation to the `.deprecated` folder is likely needed)

> even though we now know if MAR03783 has been retired, there is no reference as to who/what/when/why, so in essence the core dilemma persists

no, the current action doesn't claim to solve everything, the aim is to move step forward, hopefully in the right direction.  

> the TSV files in model/ are used for id mapping, but this new column doesn't quite have the same purpose

I don't get your point, please clarify what do you mean

> for accuracy, shouldn't one look back in past PRs to populate this new column?

agree

mihai-sysbio:
> > what is the meaning that for a given reaction the same entries HMR_3784;HMR_3783 exist in two columns rxnHMR2ID and rxnRetired?
> 
> no immediate answer, but this can definitely be figured out given the radically transparent history of data/code in Human-GEM (an investigation to the `.deprecated` folder is likely needed)

To me, the purpose of the newly added column should be clear - almost self-explanatory even to a new user. This duplication of data to me signals that the purpose is not clear enough.

> > even though we now know if MAR03783 has been retired, there is no reference as to who/what/when/why, so in essence the core dilemma persists
> 
> no, the current action doesn't claim to solve everything, the aim is to move one step forward, and hopefully in the right direction.

The proposed solution is indeed a step forward. The question is, should it end up immediately in `develop`? If it does, it will be in `main` with the next release. In my view, more steps forward are needed before this ends up in `main`. Since none defined as of now, my recommendation is for this PR to be kept in draft mode until a more robust solution is found.

> > the TSV files in model/ are used for id mapping, but this new column doesn't quite have the same purpose
> 
> I don't get your point, please clarify what do you mean

As a general logic, the TSV files contain a mapping between Human-GEM ids and ids in other databases/models. The newly introduced column is not following this logic, so I am questioning if it really belongs in this file.

haowang-bioinfo:
it should be noted that there is no new column added so far @mihai-sysbio

mihai-sysbio:
> it should be noted that there is no new column added so far @mihai-sysbio

Indeed! Sorry about the mistake, I relied on the online diff too much - it makes it look like the new column is called `rx`
Screenshot 2023-06-23 at 19 24 34

haowang-bioinfo:
>>> what is the meaning that for a given reaction the same entries HMR_3784;HMR_3783 exist in two columns rxnHMR2ID and rxnRetired?
>> 
>> no immediate answer, but this can definitely be figured out given the radically transparent history of data/code in Human-GEM (an investigation to the `.deprecated` folder is likely needed)
>
> To me, the purpose of the newly added column should be clear - almost self-explanatory even to a new user. This duplication of data to me signals that the purpose is not clear enough.

now figured out that the introduction of column `rxnRetired` was discussed in #265 and implemented #270, it basically holds the original HMR ids, therefore appears to be duplicate to `rxnHMR2ID` column

>>> even though we now know if MAR03783 has been retired, there is no reference as to who/what/when/why, so in essence the core dilemma persists
>>
>> no, the current action doesn't claim to solve everything, the aim is to move one step forward, and hopefully in the right direction.
>
> The proposed solution is indeed a step forward. The question is, should it end up immediately in `develop`? If it does, it will be in `main` with the next release. In my view, more steps forward are needed before this ends up in `main`. Since none defined as of now, my recommendation is for this PR to be kept in draft mode until a more robust solution is found (edit: by continuing the work in this PR).

both columns `rxnHMR2ID` and `rxnRetired` already existed in `main` since [v1.8.0](https://github.com/SysBioChalmers/Human-GEM/releases/tag/v1.8.0), so the current step should be fine

>>> the TSV files in model/ are used for id mapping, but this new column doesn't quite have the same purpose
>> 
>> I don't get your point, please clarify what do you mean
>
> As a general logic, the TSV files contain a mapping between Human-GEM ids and ids in other databases/models. The newly introduced column is not following this logic, so I am questioning if it really belongs in this file.

again there is no newly introduced column. Regarding the duplication between columns `rxnRetired` and `rxnHMR2ID`, the latter probably can be deprecated at some point while the former may become more useful once populated over time

haowang-bioinfo:
based on discussions on #615, from this PR (#666) and #663 onward, we started to document removed duplicate reactions and metabolites by appending their IDs to the rows of kept ones in the columns `rxnRetired` and `metRetired`, respectively
- examples can be found in either in #666 and #663
- new PRs should follow this practice as much as possible
- previously removed duplications would be revisited later
- more actions (e.g. new column) may be followed as suggested in #615

mihai-sysbio:
> This PR improves documentation of deprecated reactions in #613 by adding their ids to corresponding rows in `reactions.tsv`, as a tentative implementation according to the discussion in #615

The discussion about the implementation has yet to lead to a concrete description of the solution (= an issue). In my view tentative implementations have their place in feature branches, not in _develop_.

> both columns `rxnHMR2ID` and `rxnRetired` already existed in `main` since [v1.8.0](https://github.com/SysBioChalmers/Human-GEM/releases/tag/v1.8.0), so the current step should be fine

I see, yet I want to remember that this `rxnRetired` exists as consequence of the adoption of MA ids. If so, then to me this feels like a repurposing of that column, the vision for which was to be deleted.

> * new PRs should follow this practice as much as possible

There is no documentation of this practice in CONTRIBUTING.md, and we cannot expect new users to randomly stumble upon the discussion/PR.

haowang-bioinfo:
>> This PR improves documentation of deprecated reactions in #613 by adding their ids to corresponding rows in `reactions.tsv`, as a tentative implementation according to the discussion in #615
>
> The discussion about the implementation has yet to lead to a concrete description of the solution (= an issue). In my view tentative implementations have their place in feature branches, not in _develop_.

In my view, tentative implementation can be merged into _develop_ if discussed (#615) and approved (#666). Of course, discussion and further change is still open as always

>> both columns `rxnHMR2ID` and `rxnRetired` already existed in `main` since [v1.8.0](https://github.com/SysBioChalmers/Human-GEM/releases/tag/v1.8.0), so the current step should be fine
>
> I see, yet I want to remember that this `rxnRetired` exists as consequence of the adoption of MA ids. If so, then to me this feels like a repurposing of that column, the vision for which was to be deleted.

`rxnRetired` (`metRetired` for mets) does exist as consequence of the adoption of MA ids, and now repurposed to document removed duplicate ids. This appears to be a good solution, what do you think?

>> * new PRs should follow this practice as much as possible
>
> There is no documentation of this practice in CONTRIBUTING.md, and we cannot expect new users to randomly stumble upon the discussion/PR.

good point - go ahead to modify CONTRIBUTING.md

mihai-sysbio:
> good point - go ahead to modify CONTRIBUTING.md please

Sorry, I can't do that, as I wouldn't fully understand how the different columns are meant to be used.

Issue #639, opened 2023-06-13
FAD/Oxygen-Peroxide Duplicate Reactions


Devlin-Moyer:
## Problems

Similarly to #607, I found a number of pairs of reactions that are largely identical except one reduces FAD to FADH<sub>2</sub> and the other reduces O<sub>2</sub> to H<sub>2</sub>O<sub>2</sub>:

Reactant | FAD-dependent ID | O<sub>2</sub>-dependent ID
---|---|---
25(S)THCA-CoA | MAR01638 | MAR01639
linoleoyl-CoA | MAR02804 | MAR03301
cis,cis-palmito-7,10-dienoyl-CoA | MAR02808 | MAR03306
cis,cis-myristo-5,8-dienoyl-CoA | MAR02810 | MAR03311
4-cis-decenoyl-CoA | MAR02818 | MAR04966
palmitoleoyl-CoA | MAR03276 | MAR03364
(7Z)-tetradecenoyl-CoA | MAR03289 | MAR03369
(5Z)-dodecenoyl-CoA | MAR03295 | none

There's also this strange case:
`MAR00059: 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + FADH2 [x] + 0.5 O2 [x] ⇒ (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x] + FAD [x], GPR: ACOX2`

## Background/Rationale

ACOX2 irreversibly _oxidizes_ its substrates (see [Uniprot](https://www.uniprot.org/uniprotkb/Q99424/entry) and [this paper](https://doi.org/10.1016/j.jhep.2016.11.005)), so `MAR00059` definitely doesn't represent a reaction catalyzed by ACOX2 (see #634 for details and examples of reactions that ACOX2 should be associated with). However, `MAR00059` is very similar to `MAR01706`:

`MAR00059: 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + FADH2 [x] + 0.5 O2 [x] ⇒ (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x] + FAD [x], GPR: ACOX2`
[...]

Issue #634, opened 2023-06-12
Many Mitochondrial Beta-Oxidation Reactions Are Associated With Peroxisomal Genes


Devlin-Moyer:
As briefly mentioned in #607, some mitochondrial reactions in Human-GEM are associated with genes that catalyze analogous reactions in peroxisomes and not mitochondria. I found a total of 128 reactions that involve at least one metabolite in [m] and at least one enzyme known to be involved in fatty-acid oxidation in peroxisomes and not mitochondria:

Reaction ID | Reaction | Reaction Type | Current GPR | Proposed New GPR
---|---|---|---|---
MAR03212 | FAD + propanoyl-CoA --> acrylyl-CoA + FADH2 | Acyl-CoA Dehydrogenation | ACAD10 or ACADM or ACADS or ACAD8 or ACAD9 or ACADSB or ACAD11 | ACADS or ACADSB or ACAD8
MAR03163 | butyryl-CoA + FAD --> crotonyl-CoA + FADH2 | Acyl-CoA Dehydrogenation | ACAD10 or ACADM or ACADS or ACAD8 or ACAD9 or ACADSB or ACAD11 | ACADS or ACADSB or ACAD8
MAR02035 | palmitoyl-CoA + ubiquinone --> (2E)-hexadecenoyl-CoA + ubiquinol | Acyl-CoA Dehydrogenation | ACADVL or ACOX3 or ACADL or ACADM or ACOX1 | N/A; would be removed by #613
MAR02366 | butyryl-CoA + ubiquinone --> crotonyl-CoA + ubiquinol | Acyl-CoA Dehydrogenation | ACOX3 or ACOX1 | N/A; would be removed by #613
MAR02367 | lauroyl-CoA + ubiquinone --> (2E)-dodecenoyl-CoA + ubiquinol | Acyl-CoA Dehydrogenation | ACOX3 or ACOX1 | N/A; would be removed by #613
MAR02369 | octanoyl-CoA + ubiquinone --> (2E)-octenoyl-CoA + ubiquinol | Acyl-CoA Dehydrogenation | ACOX3 or ACOX1 | N/A; would be removed by #613
MAR02370 | myristoyl-CoA + ubiquinone --> (2E)-tetradecenoyl-CoA + ubiquinol | Acyl-CoA Dehydrogenation | ACOX3 or ACOX1 | N/A; would be removed by #613
[...]

PR #670, opened 2023-06-27
Fix FAD/Oxygen-Peroxide Duplicate Reactions

Devlin-Moyer:
As proposed in #639:
- Removes `MAR00059` for being a less accurate version of `MAR01706`
- Changes the GPR of `MAR01706` to `ENSG00000113790 or ENSG00000133835` and adds `PMID:19357427` to its references
- Removes `MAR01639` for being redundant with `MAR01638`
- Removes `MAR03301` for being redundant with `MAR02804`
- Removes `MAR03306` for being redundant with `MAR02808`
- Removes `MAR03311` for being redundant with `MAR02810`
- Removes `MAR04966` for being redundant with `MAR02818`
- Removes `MAR03364` for being redundant with `MAR03276`
- Removes `MAR03369` for being redundant with `MAR03289`
- Replaces `MAM02630x` (O<sub>2</sub>) with `MAM01802x` (FAD) and `MAM02041x` (H<sub>2</sub>O<sub>2</sub>) with `MAM01803x` (FADH<sub>2</sub>) in `MAR00442, MAR00759, MAR00777, MAR00800, MAR00833, MAR00872, MAR00952, MAR00968, MAR00974, MAR00975, MAR00977, MAR00978, MAR00982, MAR01001, MAR01005, MAR01018, MAR01162, MAR01205, MAR01254, MAR01271, MAR01291, MAR01296, MAR01303, MAR01600, MAR01704, MAR02408, MAR02510, MAR03053, MAR03054, MAR03055, MAR03056, MAR03057, MAR03062, MAR03066, MAR03070, MAR03074, MAR03078, MAR03082, MAR03086, MAR03090, MAR03094, MAR03098, MAR03102, MAR03326, MAR03330, MAR03334, MAR03338, MAR03342, MAR03346, MAR03350, MAR03356, MAR03360, MAR03459, MAR03466, MAR03470, MAR03498, MAR03506, MAR03512, MAR03849, MAR03984, MAR04001, MAR04650, MAR04789, MAR04981, MAR05025, MAR05026, MAR05028, MAR05097, MAR05100, MAR05108, MAR05119, MAR05175, MAR05178, MAR05180, MAR05182, MAR05184, MAR05189, MAR05196, MAR05300, MAR05327, MAR05331, MAR05350, MAR05366, MAR05372, MAR05374, MAR05377, MAR05382, MAR05421, MAR05428, MAR05431, MAR06284, MAR06285, MAR06322, MAR06344, MAR06345, MAR06933, MAR06936, MAR06943, MAR06946, MAR06949, MAR06952, MAR06959, MAR06962, MAR06964, MAR07641, MAR07647, MAR07704, MAR07706, MAR08017, MAR08021, MAR08415, MAR08689, MAR08690, MAR08755, MAR11529, MAR11534, MAR11539, MAR12372, MAR12808`, and `MAR20066`
- Creates new reaction `MAR20114` to represent reduction of O<sub>2</sub> by FADH<sub>2</sub>: `MAM02630x + MAM01803x -> MAM02041x + MAM01802x`, GPR: `ENSG00000161533 or ENSG00000168306 or ENSG00000087008 or ENSG00000110887 or ENSG00000203797 or ENSG00000007171 or ENSG00000148832 or ENSG00000179761 or ENSG00000158125`, reference: `PMID:30378035`


haowang-bioinfo:
> when I initially made this PR, it passed all of the metabolic task checks, but when I merged in the latest develop branch just now, it failed the "verification" tasks, but the error message doesn't give me enough details to follow up on what exactly is causing that issue, so I'm not sure how to proceed. (no rush to answer, since I know I've opened a bunch of issues and PRs recently, but just flagging that I don't know how to respond to this test failing)

will manually check what specific task(s) are affected by this PR, which appears to involve over amount of changes

haowang-bioinfo:
manual check located that the step of "reverse MAR00059 and remove 0.5 O2" (commit 9d7b81c171130f7a20c212961588d8ee1a098d44) caused the failure of [verification tasks](https://github.com/SysBioChalmers/Human-GEM/blob/main/data/metabolicTasks/metabolicTasks_VerifyModel.txt)

JonathanRob:
Good catch, @haowang-bioinfo.

The removal of the 0.5 O2 in that reaction resulted in it becoming mass-imbalanced, and reversing the reaction led to a thermodynamically infeasible reaction.

For example, a loop could be formed with reaction MAR01706:

| Rxn | Equation | 
| --- | --- |
| MAR01706 | 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + H2O [x] ⇒ (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x] |
| MAR00059 | (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x] + FAD [x] ⇒ 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + FADH2 [x] |
| Net | H20 + FAD -> FADH2 |
| Net (w/ O2) | H20 + FAD -> FADH2 + 0.5 O2 |

Even without O2, it allows water to reduce FAD, which is incorrect. With O2, it allows the same reduction, but also produces O2 from water, which is also problematic.

@Devlin-Moyer, what was your justification for reversing and removing O2 from MAR00059? It seems to me that neither of those changes should be made.

Devlin-Moyer:
From #639:
> ACOX2 irreversibly _oxidizes_ its substrates (see [Uniprot](https://www.uniprot.org/uniprotkb/Q99424/entry) and [this paper](https://doi.org/10.1016/j.jhep.2016.11.005)), so `MAR00059` should point in the opposite direction, and it should either not involve O<sub>2</sub> at all, or involve no FAD or FADH<sub>2</sub> and consume 1 O<sub>2</sub> and produce 1 H<sub>2</sub>O<sub>2</sub>, as in the other three reactions ACOX2 is currently associated with (`MAR01296`, `MAR01704`, and `MAR04001`)

I was just thinking about redox when putting together this issue, and since every other reaction on this list was straightforward redox, it never occurred to me to check the empirical formulas of these metabolites to figure out if the 0.5 O<sub>2</sub> was necessary; I stand by my initial claim the original form of MAR00059 is not a reaction that ACOX2 catalyzes, but I don't know which gene it should be associated with instead or if there are different modifications that should be made to its stoichiometry

Devlin-Moyer:
Hmm apparently if you remove the 0.5 O<sub>2</sub> from the original reaction, it exactly matches [R04547](https://www.genome.jp/entry/R04547), but KEGG doesn't associate that with any human enzymes, so do we know if this reaction happens in humans at all? Does removing it create a gap?

Devlin-Moyer:
I also found (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA in the [primary bile acid biosynthesis KEGG pathway](https://www.genome.jp/pathway/map00120) (near the bottom left, directly above chenodeoxycholate), where it participates in two reactions already represented by MAR01706 and MAR01708.

Now that I look at MAR01706 and MAR00059 (as it originally was) side-by-side:

MAR00059: 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + FADH2 [x] + 0.5 O2 [x] ⇒ (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x] + FAD [x]
MAR01706: 3alpha,7alpha-dihydroxy-5beta-cholest-24-enoyl-CoA [x] + H2O [x] ⇒ (24R,25R)3alpha,7alpha,24-trihydroxy-5beta-cholestanoyl-CoA [x]

I think the situation is that MAR00059 is a less-accurate version of MAR01706, maybe added during gap filling at some point before MAR01706 existed, so probably MAR00059 should just be removed altogether as a (near-)duplicate of MAR01706. Let me know if you think that checks out and I can update everything accordingly

haowang-bioinfo:
> Hmm apparently if you remove the 0.5 O2 from the original reaction, it exactly matches [R04547](https://www.genome.jp/entry/R04547), but KEGG doesn't associate that with any human enzymes, so do we know if this reaction happens in humans at all? Does removing it create a gap?

so the change made to MAR00059 is based on KEGG, which actually is known with rich amount of unbalanced reactions. [Genes](https://www.genome.jp/dbget-bin/get_linkdb?-t+genes+rn:R04547) associate with R04547 all appear to be from bacterial kingdom.

> Does removing it create a gap? Let me know if you think that checks out

it's confirmed that removal of MAR00059 could fix aforementioned failed verification tasks

JonathanRob:
Thanks for the thorough update, @Devlin-Moyer. I agree that removing MAR00059 seems like a good solution, and thank you @haowang-bioinfo for verifying.

haowang-bioinfo:
if the GPR of MAR01706 should be modified, and how?

Devlin-Moyer:
Based on the research I did in #634, the GPR of `MAR01706` should be `HADHA or ECHS1` (i.e. `ENSG00000084754 or ENSG00000127884`)
I'm currently working on updating #639 to accurately and completely describe the issues with `MAR00059`, why it should be removed, and can also throw in this change to the GPR of `MAR01706`

haowang-bioinfo (approved):

haowang-bioinfo:
@Devlin-Moyer which Rhea reaction could be associated with MAR01706?

Devlin-Moyer:
Pretty sure it's [47408/47409](https://www.rhea-db.org/rhea/47408)

haowang-bioinfo:
>> @Devlin-Moyer which Rhea reaction could be associated with MAR01706?
>>
> Pretty sure it's [47408/47409](https://www.rhea-db.org/rhea/47408)
>
would it be a good idea to update Rhea id for MAR01706 in #707?

PR #672, opened 2023-07-04
feat: add basic memote tests

### LINKED ISSUE #671: bug in test setup
#### Current behavior:

It was noticed in a [previous discussion](https://github.com/SysBioChalmers/Human-GEM/pull/635#issuecomment-1618324715) that tests pass when they should in fact not.

Tests should fail if there are typos in the expected sections of the YAML.

See the checks in [this commit](https://github.com/SysBioChalmers/Human-GEM/pull/635/commits/5f39c61e3e4d6e8d3bb425037755fda90a27b7ff).

#### Anything else:

It would be nice to have tests running with MEMOTE.

mihai-sysbio:

This PR resolves #671 by introducing a basic MEMOTE tests in the `yaml-validation` workflow, and another new test that makes sure all reactions have at least one metabolite.

- [x] Selected `fix/importExportCycle` as a target branch

**Notes**
This PR is not aimed directly at `develop`, but at another branch where a bug was artificially introduced. The aim of this PR is then to capture this failure, then merge this PR. Finally, after reverting the commit introducing the bug, the `fix/importExportCycle`  will have valid content with improved workflows.

### DIFF
--- a/.github/workflows/yaml-validation.yml
+++ b/.github/workflows/yaml-validation.yml
@@ -25,5 +25,8 @@ jobs:
           ## Return non-zero exit code on warnings as well as errors
           # strict: # optional, default is false
 
+      - name: Basic MEMOTE tests
+        run: python code/test/memoteTest.py
+
       - name: Test import with cobrapy and consistency with annotation files
         run: python code/test/sanityCheck.py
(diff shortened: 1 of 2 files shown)


haowang-bioinfo (approved):
the newly introduced memote test circumvent the pending error from typo in the `metabolites` key in yaml - this is great

another duplicate reaction check based on memote is also added in this PR, besides #582. In my view, it's okay have both

PR #563, opened 2023-05-11
feat: add hart2015 essentiality suite

haowang-bioinfo:
#### Main improvements in this PR:
This PR provides infrastructure of essentiality test suite, as proposed in #390, using Hart2015 datasets with following components:
- `Hart2015_RNAseq.txt`: RNA-seq sequences of Hart2015 cell lines (GEO: GSE75189)
- `Hart2015_TableS2.xlsx`: Calcuated Bayes factors for CRISPR targeted genes
- `getTaskEssentialGenes`: Identify genes essential for different tasks in different tINIT models
- `estimateEssentialGenes`: Generate tINIT models and estimate essential genes
- `evaluateHart2015Essentiality`: Evaluate and compare Hart2015 experimental fitness genes with predicted results

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [ ] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

PR #675, opened 2023-07-07
feat: gene essentiality workflow

mihai-sysbio:

This PR aims to automatise the gene essentiality checks as introduced in https://github.com/SysBioChalmers/Human-GEM/pull/563 with a GitHub workflow, also discussed in that PR thread.


.github/workflows/gene-essentiality.yml
```diff
@@ -0,0 +1,32 @@
...
+    runs-on: self-hosted
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Run gene essentiality
+        id: essentiality
+        run: |
+          TEST_RESULTS=$(/usr/local/bin/matlab -nodisplay -nosplash -nodesktop -r "ihuman = importYaml('Human-GEM.yml'); taskStruct = parseTaskList('data/metabolicTasks/metabolicTasks_Essential.txt'); eGenes = estimateEssentialGenes(ihuman, 'Hart2015_RNAseq.txt', taskStruct); disp(evaluateHart2015Essentiality(eGenes));" | awk 'NR>9 && !/^\.+/')
```
mihai-sysbio, line 19: @haowang-bioinfo the output of this line is very hard to capture/use. Do you have any suggestions to make this output more compact?

mihai-sysbio:
Also noting here this relevant research article _[Synthetic lethality in large-scale integrated metabolic and regulatory network models of human cells](https://www.nature.com/articles/s41540-023-00296-3)_.

haowang-bioinfo:
> Also noting here this relevant research article [Synthetic lethality in large-scale integrated metabolic and regulatory network models of human cells](https://www.nature.com/articles/s41540-023-00296-3).

yes, it mentioned another method [gMCStool](https://www.biorxiv.org/content/10.1101/2022.11.03.514827v1) that can predict essential genes in context-specific models

.github/workflows/gene-essentiality.yml
```diff
@@ -0,0 +1,32 @@
+name: Check gene essentiality with Hart 2015
+
+on:
+  pull_request:
+    branches:
+      - "main"
```
mihai-sysbio, line 6: It might be safer to start by allowing this workflow to run also in PRs targeting `develop`:
```suggestion
      - "main"
      - "develop"
```

.github/workflows/gene-essentiality.yml
mihai-sysbio: As seen [in this workflow run](https://github.com/SysBioChalmers/Human-GEM/actions/runs/6429997344/job/17460119226?pr=675), the output is way too rich to fit in a PR comment. @haowang-bioinfo how would you suggest to change it to end up with a nice summary like [here](https://github.com/SysBioChalmers/Human-GEM/pull/700#issuecomment-1738614979)?

feiranl:
@mihai-sysbio Could you explain the reason for the failing task?

mihai-sysbio:
> @mihai-sysbio Could you explain the reason for the failing task?

The task isn't actually failing, really. As mentioned in the [comment above](https://github.com/SysBioChalmers/Human-GEM/pull/675#discussion_r1348603333), it's simply that the output of the gene essentiality is so long that it cannot be posted as a PR comment. I couldn't figure out how @haowang-bioinfo ran the calculations to obtain a much more compressed output. As visible in the workflow changes, the code here is just calling functions defined elsewhere.

haowang-bioinfo:
> I couldn't figure out how @haowang-bioinfo ran the calculations to obtain a much more compressed output. As visible in the workflow changes, the code here is just calling functions defined elsewhere.

the concise output was an outcome of a series of manual operations of running functions, copy/paste results, and modifications with text editor

.github/workflows/gene-essentiality.yml
haowang-bioinfo: no, I've never thought about this. Because they are non-binary indicators that are hard to be used for evaluating purpose in actions

mihai-sysbio:
If these last manual steps could be ran as a function, then we would be ready to run the gene essentiality analysis on every PR 💪🏻

edkerk:
As is suggested for MACAW results in #840, a summary output could be posted on in the PR, while a more detailed text file could be written and committed as part of each PR. This could be organized in the same folder as the MACAW results, and a README.md file could indicate for both output files from which PR they derived from.

edkerk:
Finally 😄 it seems to work! Ready for review.

mihai-sysbio (commented):
As a contributor to this PR I cannot formally review it - only comments are allowed - so please consider this an approval 👍🏻

Issue #673, opened 2023-07-05
Citrate Synthase Is Not Reversible In Humans


Devlin-Moyer:
`MAR04145: CoA [m] + H+ [m] + citrate [m] ⇔ H2O [m] + OAA [m] + acetyl-CoA [m], GPR: ENSG00000062485` ([citrate synthase](https://www.uniprot.org/uniprotkb/O75390/entry)) should be irreversible, since the human citrate synthase enzyme (and apparently [most organisms' citrate synthases](https://doi.org/10.1126/science.aao2410)) only catalyze the forward reaction (source focused on the human enzyme specifically is page 130 of [this book](https://books.google.com/books?hl=en&lr=&id=A3u1891gHfkC&oi=fnd&pg=PP1&ots=Ap4zUyLCPY&sig=hw2H0Zb-04QqkQ4StEMr3aX8DI4#v=onepage&q=citrate&f=false))

feiranl:
Nice catch! Please go ahead and make the change

Devlin-Moyer:
Resolved by #676

PR #676, opened 2023-07-16
Make Citrate Synthase Irreversible

Devlin-Moyer:

Makes `MAR04145`, the citrate synthase reaction, irreversible, as proposed in #673
Also swapped the products and reactants so that it could only have non-negative fluxes instead of non-positive fluxes
Also added information about the textbook referenced in #673 to the rxnNotes, since that seems like the way references without PMIDs are handled (let me know if I should've done that differently)


feiranl (approved):
LGTM!

PR #692, opened 2023-08-16
fix: refine animal GEM updating pipeline

haowang-bioinfo:
This PR improves animal GEM updating pipeline:
  - sync `geneShortNames` field (introduced by #539) by padding with empty elements in function `addMetabolicNetwork`
  - enable `updateAnimalGEM` with option of reusing Human-GEM biomass equation

### DIFF
--- a/code/updateAnimalGEM.m
+++ b/code/updateAnimalGEM.m
@@ -84,6 +84,7 @@ end
 
 %% Gap-filling
 [animalGEM, gapfillNetwork]=gapfill4EssentialTasks(animalGEM,ihuman,resetBiomass);
+animalGEM.b = animalGEM.b(:,1);   % ensure b field in single column
 
 
 %% post-gapfilling procedures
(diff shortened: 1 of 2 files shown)


code/updateAnimalGEM.m
```diff
@@ -11,7 +11,9 @@
 %   rxnsToAdd            the structure of species-specific reactions
 %   metsToAdd            the structure of species-specific metabolites
 %   modelId              model id
-%
+%   resetBiomass         reset biomass objective function to "biomass_componenets"
```
mihai-sysbio, line 14:
```suggestion
%   resetBiomass         reset biomass objective function to "biomass_components"
```

mihai-sysbio: Interesting choice here. The way I pictured this, when updating an animal GEM the biomass (almost?) always need to be reset, so there wouldn't need to be a parameter for this.

haowang-bioinfo: reset biomass is the default. Here provides alternative of reusing HumanGEM biomass and may be better option for mammals

mihai-sysbio: That's exactly my point. I think the choice of the objective function should be reset. And since this is part of a script that is only meant to be run automatically, it will be reset with every update of the animal GEMs.

At the same time, to facilitate the choice of reusing the biomass objective in different projects, I would encourage to have a new page under Human-GEM guide, e.g. "Modifications for animal GEMs", in which applying the choice of the objective function is clearly explained. There we could also have a section about the gene identifiers being different in the models.

haowang-bioinfo: not sure if it is only meant for automated pipeline, because user may want to use it separately

mihai-sysbio (approved):

PR #696, opened 2023-08-29
fix: update name for GAP in CellfieConsensus metabolic tasks

exaexa:
This is a simple fix that follows the previous commits (https://github.com/exaexa/Human-GEM/commit/8182ade485807697c2e79cf12b2a89f91e5f20cb) to remove old name for GAP[c].

### DIFF
--- a/data/metabolicTasks/metabolicTasks_CellfieConsensus.txt
+++ b/data/metabolicTasks/metabolicTasks_CellfieConsensus.txt
@@ -211,7 +211,7 @@
 				NAD+[m]	1	1	CO2[m]	1	1								
 							NADH[m]	1	1								
 	34	Synthesis of fructose-6-phosphate from erythrose-4-phosphate (HMP shunt)		fructose-6-phosphate[c]	1	1	erythrose-4-phosphate[c]	1	1								
-				GAP[c]	1	1	D-xylulose-5-phosphate[c]	1	1								
+				D-glyceraldehyde 3-phosphate[c]	1	1	D-xylulose-5-phosphate[c]	1	1								
 	35	Synthesis of ribose-5-phosphate		glucose[c]	1	1	ribose-5-phosphate[c]	1	1								
 				ATP[c]	1	1	ADP[c]	1	1								
 				NADP+[c]	2	2	H+[c]	3	3								
@@ -226,7 +226,7 @@
 	38	Glycogen degradation		glycogenin G4G7[c]	1	1	glucose[c]	8									
 				Pi[c]	3		NA[c]	1	1								
 				H2O[c]	8		glucose-1-phosphate[c]	3									
-	39	Fructose degradation (to glucose-3-phosphate)		fructose[c]	1	1	GAP[c]	2	2								
+	39	Fructose degradation (to glucose-3-phosphate)		fructose[c]	1	1	D-glyceraldehyde 3-phosphate[c]	2	2								
 				ATP[c]	2	2	ADP[c]	2	2								
 							H+[c]	2	2								
 	40	Fructose to glucose conversion (via fructose-6-phosphate)		fructose[c]	1	1	glucose[c]	1	1								


exaexa:
Extra comment: I can't find the other "missing" metabolite, `NA[c]`. Is there any way how to find these metabolites systematically?

mihai-sysbio:
> This is a simple fix that follows the previous commits ([exaexa@8182ade](https://github.com/exaexa/Human-GEM/commit/8182ade485807697c2e79cf12b2a89f91e5f20cb)) to remove old name for GAP[c].

Thanks @exaexa. I think I've seen this recently 🤔

exaexa:
(cc @migp11 )

haowang-bioinfo (approved):

exaexa:
@haowang-bioinfo @mihai-sysbio is there please any info on the `NA[c]` metabolite? ref: https://github.com/SysBioChalmers/Human-GEM/blob/e16451ec2766f40590f3543551a77dd8f530cee0/data/metabolicTasks/metabolicTasks_CellfieConsensus.txt#L223  (and several other places)

JonathanRob:
@exaexa these metabolic tasks were imported/translated from their source at [Cellfie](https://github.com/LewisLabUCSD/CellFie), seemingly from [CONSENSUS_TASKS.xls](https://github.com/LewisLabUCSD/CellFie/blob/master/input/CONSENSUS_TASKS.xls).

I checked the corresponding task in this original source file, and it looks like the `NA[c]` metabolite should be [`Tyr_ggn[c]`](http://bigg.ucsd.edu/universal/metabolites/Tyr_ggn). It is unclear to me what metabolite this should correspond to in Human-GEM - maybe [`glycogenin[c]`](https://metabolicatlas.org/explore/Human-GEM/gem-browser/metabolite/MAM01996c)?

exaexa:
Thanks a lot! I'll have a look asap and hopefully send another PR today

migp11:
Hi,

I've been checking for [Tyr_ggn_e](http://bigg.ucsd.edu/models/Recon3D/metabolites/Tyr_ggn_e) in BiGG and it says that this metabolite is _Tyr-194 of apo-glycogenin protein (primer for glycogen synthesis)_. Furthermore, The same cross-reference is found in [VHM](https://www.vmh.life/#metabolite/Tyr_ggn).

Then I followed the cross-reference to MetaNetX through the identifier [MNXM146651](https://www.metanetx.org/chem_info/MNXM146651) and it seems this ID is deprecated and replaced by [MNXM728887](https://www.metanetx.org/chem_info/MNXM728887). 

The strange thing is that in MetaNetX  MNXM728887 corresponds to _Thyroglobulin L-tyrosine_ and after a quick search I found that Thyroglobulin is not involved in glycogen metabolism but instead, Thyroglobulin is a 660 [kDa](https://en.wikipedia.org/wiki/KDa), [dimeric](https://en.wikipedia.org/wiki/Dimer_(chemistry)) [glycoprotein](https://en.wikipedia.org/wiki/Glycoprotein) produced by the [follicular cells](https://en.wikipedia.org/wiki/Thyroid_epithelial_cell) of the [thyroid](https://en.wikipedia.org/wiki/Thyroid) and used entirely within the thyroid gland [[reference](https://en.wikipedia.org/wiki/Thyroglobulin)]. 

Therefore, my conclusion is that the cross-reference to MetNetX is wrong. Tyr_ggn is "Tyr-194 Of Apo-Glycogenin Protein" the glycoprotein involved in glycogen biosynthesis.

Finally, I checked reactions in Recon3D (VHM) for this metabolite and the only relevant (not an exchange/source/sink) is the following:

[GLPASE2](https://www.vmh.life/#reaction/GLPASE2)
Glycogen Phosphorylase (Amyls -> Glc-D)
glygn3[c] + 7.0 h2o[c] -> 7.0 glc_D[c] + Tyr_ggn[c]

and although it is cross-referenced to MetaNetX through ([MNXR100266](https://www.metanetx.org/equa_info/MNXR100266)) I was not able to find the corresponding reactions in HumanGEM. Thus, to avoid other users running into the same issue, this task should be marked as conflicting with HumanGEM or some kind of warning.

My two cents!
Best
Miguel

haowang-bioinfo:
@migp11 such a nice and comprehensive investigation!

totally agree that this task need to be disabled, or at least warned in checking tasks.

exaexa:
@migp11 oh whoops I missed your investigation before I started #697... can you pls check if the fix makes sense at least partially there? :]

haowang-bioinfo:
one probable solution is set feature `SHOULD FAIL` of this row to "TRUE", and mark in `COMMENTS` column that: "This task is conflicting with HumanGEM"

migp11:
@haowang-bioinfo, your suggestion might be a good workaround, or not, depending on how missing/unknown metabolites are treated when a task is created. If unknown metabolites are ignored then, it is possible that the task would be feasible (though biochemically inconsistent). On the other hand, if an error is raised when an unknown metabolite is found on a task, then it is clear that the task should not be used.

migp11:
One last though, seen NA[c] for several different metabolites makes me think that NA might mean "Not Available" and it is added whenever a metabolite in a task is not present in the model

mihai-sysbio:
Pitching in as well, I wouldn't mind seeing the link to this issue in the COMMENTS field.

haowang-bioinfo:
> I believe one can "comment-out" rows in the task file by adding a # in the first column. So a solution could be to comment-out the rows corresponding to the problematic task, while also adding a comment as @haowang-bioinfo suggests. But I haven't tested adding a # since we changed the task file format from .xlsx to .tsv, so I'm not sure if it will work.

adding hash tag seems to be a good option, and works well by testing with the latest `.tsv` file

haowang-bioinfo:
> indeed, multiple occurrence of NA[c] can be observed, such as in tasks number 37, 38, and 181. No sure if NA[c] represents "Not Available", could it be something like a place holder?

to move #697 forward, would lean toward adding hash tag to comment-out tasks 37, 38, and 181

Issue #599, opened 2023-05-30
Proposed fixes to DHAP <=> sn-glycerol-3-phosphate reactions


jreimertz:
I recently came across five reactions that are involved in the conversion of DHAP to sn-glycerol-3-phosphate in glycolysis that I've included below:

| Reaction ID | Reaction String                                                                                                  | GPR               |
| ------------- | ----------------------------------------------------------------------------------- | -------------- | 
| MAR01169  | FAD [m] + sn-glycerol-3-phosphate [c] ⇒ DHAP [c] + FADH2 [m]                   | GPD2             |
| MAR00483  | sn-glycerol-3-phosphate [c] + ubiquinone [m] ⇒ DHAP [c] + ubiquinol [m]   | GPD2             |
| MAR00482  | FAD [c] + sn-glycerol-3-phosphate [c] ⇒ DHAP [c] + FADH2 [c]                      | GPD2             |
| MAR00479  | DHAP [c] + H+ [c] + NADH [c] ⇒ NAD+ [c] + sn-glycerol-3-phosphate [c]     | GPD1L, GPD1 |
| MAR00478  | DHAP [c] + H+ [c] + NADPH [c] ⇒ NADP+ [c] + sn-glycerol-3-phosphate [c] | GPD1             |

The Uniprot page for [GPD2](https://www.uniprot.org/uniprotkb/P43304/entry) supports the reaction MAR00483, but contradicts MAR00482 and MAR01169. Similarly, the Uniprot pages for [GPD1](https://www.uniprot.org/uniprotkb/P21695/entry) and [GPD1L ](https://www.uniprot.org/uniprotkb/Q8N335/entry) support MAR00479 but don't include MAR00478. Additionally, I wasn't able to find any papers that mentioned NADPH/NADP+ in this reaction in humans.

As a result, I think it would be appropriate to remove reactions MAR00482, MAR00478, and MAR01169

feiranl:
Good catch! 
I am ok with removing MAR00478 and MAR00482.
I am not sure about this. I am not familiar with this protein, just from Uniprot annotation, it appears to have FAD binding region. 

> Cofactor
> Protein has 1 cofactor binding site:
[...]

PR #710, opened 2023-10-17
Fix Duplicate Glycerol-3-Phosphate Dehydrogenase Reactions

Devlin-Moyer:

As proposed at [the end of the discussion](https://github.com/SysBioChalmers/Human-GEM/issues/599#issuecomment-1599513619) in #599:
- Removes `MAR00478` because there's no evidence that GPD1 or GPD1L can use NADP(H) in addition to NAD(H)
- Removes `MAR00482` because GPD2 is definitely localized to the inner mitochondrial membrane, so it should use FAD(H2) [m] and not [c]

Also while not explicitly proposed in #599:
- Merges annotations for `MAR00478` in `reactions.tsv` and its associated reference with those for `MAR00479`, since they're nearly identical reactions
- Merges annotations for `MAR00482` in `reactions.tsv` into those for `MAR01169`, since, again, they're nearly identical reactions


feiranl (approved):
LGTM!

haowang-bioinfo (approved):

Issue #755, opened 2023-12-10
Localization of DECR1 and DECR2


Devlin-Moyer:
### Current behavior:

[DECR1](https://www.uniprot.org/uniprotkb/Q16698/entry#subcellular_location) and [DECR2](https://www.uniprot.org/uniprotkb/Q9NUI1/entry) both catalyze NADPH-dependent reduction of 2,4-dienoyl-CoAs during beta-oxidation, but DECR1 does so exclsively in mitochondria and DECR2 does so exclusively in peroxisomes. Several reactions in Human-GEM have GPRs of "DECR1 or DECR2"

ID|Reaction
---|---
`MAR01176` | 8(R)-hydroxy-hexadeca-(2E,4E,6E,10Z)-tetraenoate [m] + H+ [m] + NADPH [m] --> 8(R)-hydroxy-hexadeca-(2E,6E,10Z)-trienoate [m] + NADP+ [m]
`MAR01181` | 6(R)-hydroxy-tetradeca-(2E,4E,8Z)-trienoate [m] + H+ [m] + NADPH [m] <=> 6(R)-hydroxy-tetradeca-(2E,8Z)-dienoate [m] + NADP+ [m]
`MAR01218` | 8(S)-hydroxy-hexadeca-(2E,4E,6E,10Z)-tetraenoate [m] + H+ [m] + NADPH [m] --> 8(S)-hydroxy-hexadeca-(2E,6E,10Z)-trienoate [m] + NADP+ [m]
`MAR01222` | 6(S)-hydroxy-tetradeca-(2E,4E,8Z)-trienoate [m] + H+ [m] + NADPH [m] <=> 6(S)-hydroxy-tetradeca-(2E,8Z)-dienoate [m] + NADP+ [m]
~~`MAR03296`~~ | 2-trans-4-cis-decadienoyl-CoA [m] + H+ [m] + NADH [m] --> NAD+ [m] + trans-3-decenoyl-CoA [m]
`MAR03446` | 2-trans-cis,cis,cis,cis-4,8,11,14-eicosapentaenoyl-CoA [m] + H+ [m] + NADPH [m] --> NADP+ [m] + trans-3-cis-8,11,14-eicosatetraenoyl-CoA [m]
[...]

Issue #736, opened 2023-12-03
DECR1 and DECR2 use NADPH, not NADH


Devlin-Moyer:
### Current behavior:

These are all reactions currently in Human-GEM that use NADH to reduce 2,4-dienoyl-CoAs and are associated with DECR1 and/or DECR2:

ID|Reaction
---|---
`MAR03322` | 2-trans-4-cis-decadienoyl-CoA [x] + H+ [x] + NADH [x] --> NAD+ [x] + trans-3-decenoyl-CoA [x]
`MAR03296` | 2-trans-4-cis-decadienoyl-CoA [m] + H+ [m] + NADH [m] --> NAD+ [m] + trans-3-decenoyl-CoA [m]
`MAR06433` | 11-carboxy-alpha-tocotrienol [m] + ATP [m] + CoA [m] + 2 H2O [m] + NAD+ [m] + O2 [m] --> 9-carboxy-alpha-tocotrienol [m] + acetyl-CoA [m] + AMP [m] + 2 H+ [m] + H2O2 [m] + NADH [m] + PPi [m]
`MAR06448` | 11-carboxy-gamma-tocotrienol [m] + ATP [m] + CoA [m] + 2 H2O [m] + NAD+ [m] + O2 [m] --> 9-carboxy-gamma-tocotrienol [m] + acetyl-CoA [m] + AMP [m] + 2 H+ [m] + H2O2 [m] + NADH [m] + PPi [m]

### Expected behavior:

[DECR1](https://www.uniprot.org/uniprotkb/Q16698/entry) (the mitochondrial enzyme) and [DECR2](https://www.uniprot.org/uniprotkb/Q9NUI1/entry) (the peroxisomal equivalent) both appear to be exclusively capable of using NADPH and not NADH to catalyze this reaction. `MAR04970` and `MAR04974` already represent NADPH-dependent equivalents of `MAR03296` and `MAR03322`, respectively, so `MAR03322` and `MAR03296` can just be removed. The other two have a number of other problems, which I address separately in #738 

### Proposed changes:

- [ ] Remove `MAR03322` for being a less accurate version of `MAR04974`
- [ ] Remove `MAR03296` for being a less accurate version of `MAR04970`

haowang-bioinfo:
[...]

PR #788, opened 2023-12-26
fix: revise rxns and GPRs associated with DECR1 and DECR2

haowang-bioinfo:
This PR revise reactions and GPRs associated with DECR1 and DECR2
- remove MAR03322 and MAR03296 due to ~~compartment~~ cofactor mismatch, as reported in #736
- sort out GPRs according to compartment localication, as porposed in #755
  - set the GPRs of MAR01176, MAR01181, MAR01218, MAR01222, MAR03446, and MAR03457 to ENSG00000104325
  - set the GPRs of MAR01297, MAR03461, and MAR03471 to ENSG00000242612


Devlin-Moyer (approved):
All the changes look good to me; the only note I have is that in the PR message you say `MAR03322` and `MAR03296` are removed because of a "compartment mismatch", but the problem is that they used the wrong cofactor (NAD instead of NADP), not that they happened in the wrong compartment. Anyway I assume that was a typo, and since you link directly to the relevant issue where the cofactor situation is explained, it's not much of a problem.

haowang-bioinfo:
> All the changes look good to me; the only note I have is that in the PR message you say MAR03322 and MAR03296 are removed because of a "compartment mismatch", but the problem is that they used the wrong cofactor (NAD instead of NADP), not that they happened in the wrong compartment. Anyway I assume that was a typo, and since you link directly to the relevant issue where the cofactor situation is explained, it's not much of a problem.

sorry, that's my mistake and now corrected

PR #839, opened 2024-09-15
Fix/peptide hydrolysis

### LINKED ISSUE #838: Peptide metabolism reactions are reversible
#### Current behavior:
Hi all,

The reactions in the subsystem `Peptide metabolism` are all **reversible**, meaning that peptides can be formed from individual amino acids without an ATP cost. Also, I do not think that peptide synthesis even could work this way in animals.
Moreover, a couple of peptide hydrolysis reactions are assigned to other subsystems, this also needs fixing.

#### Curation plan:
1. Set the subsystem of wrongly-annotated reactions (MAR11413, MAR11414, MAR11411, MAR06898, MAR06861, MAR06915, MAR07149, MAR07193, MAR08979) to `Peptide metabolism`;
2. Set all reactions in this subsystems as irreversible.

How does that sound?

pranasag:
This PR fixes #838 by (i) setting the correct subsystem to some peptide hydrolysis reactions, and (ii) by setting all reactions in subsystem "Peptide metabolism" to irreversible. Alongside, the names of reactions have been changed to explicitly indicate they are hydrolysis reactions.

- [x] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists


JHL-452b:
Looks good. I noticed that in addition to the 9 reactions mentioned in the #838 , a large number of similar reactions were also modified. I think it would be better if you could provide a list of the entire modified reactions, so that it would be easier to trace back later.

pranasag:
These are all reactions in "Peptide metabolism" subsystem (~250) - do you really want to list them in text format here? I already highlighted that the whole subsystem had to be reviewed to proposed implementation in #838, and this comes as a separate commit.

JHL-452b:
Oh, sorry, I didn't notice that, if it's already been mentioned then there's no need to list it. 

It looks fine to me, and the implementation matchs with the claims in PR message.

edkerk (approved):

PR #883, opened 2024-10-19
chore: update model I/O functions and documentation

edkerk:
#### Main improvements in this PR:
- Refactor:
  - Use RAVEN functions `readYAMLmodel`, `writeYAMLmodel` and `exportForGit`, instead of the Human-GEM specific `importYaml`, `exportYaml` and `exportHumanGEM`.
  - Modify `increaseHumanGEMVersion`, and include a check that `Human-GEM.yml` and `model/tsv` files match.
  - Prepare for `Human-GEM.txt` to show `metNames[comps]` instead of `metIDs`.
  - Full round of I/O with `Human-GEM.yml` using RAVEN 2.10.3. Changes in `txt` (major, see above), `xml` and `mat` (both minor) will appear with next release.
  - Remove duplicate `getGenesFromGrRules` function, already present in RAVEN.

- Documentation:
  - Update `README.md` to reflect changes in model I/O functions, more clearly explain the software requirements and the ability to contribute to development without RAVEN.

**I hereby confirm that I have:**

- [x] Tested my code on my own computer for running the model
- [x] Selected `develop` as a target branch
- [ ] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

PR #906, opened 2024-11-05
fix: allow macaw and gene essentiality to run simultaneously

edkerk:
Both the MACAW and gene essentiality tests commit the PR number in `data/testResults/README.md`, but this will cause conflicts if they are mentioned in the same line.


edkerk:
Combined in #883

Issue #890, opened 2024-10-28
GPR curation for Acyl-CoA hydrolysis, Acylglycerides metabolism and Nicotinate and nicotinamide metabolism


JHL-452b:
<!-- Note: Please search to see if an issue already exists for the bug you encountered. -->

#### Current behavior:
<!-- A concise description of what you're experiencing. Try to be as clear as possible: Is it something wrong/missing in the model? Is it a simulation issue? How does the reaction/metabolite/gene/simulation look like in the `main` branch? -->
17 genes are discovered to be uncorrelated with their corresponding reactions in `Acyl-CoA hydrolysis`， `Acylglycerides metabolism` and `Nicotinate and nicotinamide metabolism`.
| Gene| UniProtID| Reaction| GPR| Correlation| Manually check |
|---------|--------------|----------|------------|-----------|----------------|
| ENSG00000136881 | [Q14032](https://www.uniprot.org/uniprotkb/Q14032/entry) | MAR00242: (10Z)-heptadecenoyl-CoA + H2O --> (10Z)-heptadecenoic acid + CoA + H+| ENSG00000184227 or ENSG00000097021 or ENSG00000136881 or ENSG00000112304 or ENSG00000159445| Pending No, the gene is not relevant to the reaction provided. Analysis and explanation. The gene ENSG00000136881 is involved in the amidation of bile acids with amino acids like taurine and glycine, as well as in acyl-CoA thioesterase activities involving long- and very long-chain acyl-CoAs. The provided reaction, which describes the hydrolysis of (10Z)-heptadecenoyl-CoA to (10Z)-heptadecenoic acid, CoA, and H+, does not align with the gene's known functions and catalytic activities. The gene's activities are specific to bile acid conjugation and the hydrolysis of saturated and other specific long-chain acyl-CoAs, while the provided reaction involves an unsaturated fatty acid, which is not mentioned as a substrate for the gene's enzymatic functions.| No|
[...]

PR #934, opened 2024-12-06
Fix: GPR curation for acyl-CoA hydrolysis and nicotinate and nicotinamide metabolism

JHL-452b:
As proposed in #890 
- Remove `ENSG00000112304` from MAR00242;
- Replace `ENSG00000205669` with `ENSG00000159445` from MAR00324;
- Replace `ENSG00000205669` with `ENSG00000159445` from MAR00332;
- Add `ENSG00000159445` for MAR00354;
- Replace `ENSG00000177465` and `ENSG00000101473` with `ENSG00000172497` and `ENSG00000159445` from MAR03009;
- Remove `ENSG00000112304` from MAR00210;
- Replace `ENSG00000205669` with `ENSG00000159445` from MAR00426;
- Replace `ENSG00000205669` with `ENSG00000159445` from MAR00438;
- Remove `ENSG00000205309`, `ENSG00000076685` from MAR04252;
- Remove `ENSG00000205309`, `ENSG00000076685` from MAR04264;
- Remove `ENSG00000162836` from MAR04270;
- Remove `ENSG00000198805` from MAR04662;
- Remove `ENSG00000109743` from MAR08788.

### DIFF
--- a/model/Human-GEM.yml
+++ b/model/Human-GEM.yml
@@ -108880,7 +108880,7 @@
           - MAM02745c: 1
       - lower_bound: 0
       - upper_bound: 1000
-      - gene_reaction_rule: "ENSG00000184227 or ENSG00000097021 or ENSG00000136881 or ENSG00000112304 or ENSG00000159445"
+      - gene_reaction_rule: "ENSG00000184227 or ENSG00000097021 or ENSG00000136881 or ENSG00000159445"
       - rxnNotes: ""
       - rxnFrom: "HMRdatabase"
       - eccodes: "3.1.2.2"
@@ -108994,7 +108994,7 @@
           - MAM02040c: -1
       - lower_bound: 0
       - upper_bound: 1000
-      - gene_reaction_rule: "ENSG00000184227 or ENSG00000097021 or ENSG00000136881 or ENSG00000112304 or ENSG00000159445"
+      - gene_reaction_rule: "ENSG00000184227 or ENSG00000097021 or ENSG00000136881 or ENSG00000159445"
       - rxnNotes: ""
       - rxnFrom: "HMRdatabase"
       - eccodes: "3.1.2.2"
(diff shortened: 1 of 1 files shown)


JHL-452b:
I'm sorry I didn't check carefully enough before. I am now realizing that the modification I proposed for MAR03009 may have a problem. 

What we could know here is that `ENSG00000172497` and `ENSG00000159445` have a hydrolyzing function for most acyl-CoAs in the cytosol. However, [MAR03009](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR03009) is located in peroxisomal rather than cytosol. This is the root cause of the problem.

The original GPR of `MAR03009` is [ENSG00000101473](https://www.uniprot.org/uniprotkb/O14734/entry) or [ENSG00000177465](https://www.uniprot.org/uniprotkb/Q8N9L9/entry). Both of them have the hydrolyzing function for acyl-CoAs and located in peroxisomal. So, I think MAR03009 should keep the original GPR. What do you think? @Devlin-Moyer

JHL-452b:
> ACOT4 (ENSG00000177465) should still be removed, because, while it is capable of hydrolyzing acyl-CoAs in peroxisomes, it is not active towards propanoyl-CoA (see figure 4A of [this paper](https://faseb.onlinelibrary.wiley.com/doi/epdf/10.1096/fj.06-6042com))

The human ACOT4(`ENSG00000177465`) possesses activity with medium- to long chain acyl-CoAs in addition to the activity with succinyl-CoA and glutaryl-CoA. But the corresponding C4-monocarboxy-lyl-CoA (butyryl-CoA) was not a substrate, and in contrast the C12 dicarboxylyl-CoA was not a substrate(even though the corresponding C12 monocarboxylyl-CoA was a good substrate). Therefore, I believe that ACOT4 is also inactive towards propanoyl-CoA (similar to butyryl-CoA) in `MAR03009`. Agree to remove ACOT4 (ENSG00000177465) for `MAR03009`.

As for ACOT12(`ENSG00000172497`), I think we should not add it to the GPR of MAR03009 for the time being. We will make a judgment after finding stronger evidence.

feiranl:
@edkerk  Could check this PR?

PR #998, opened 2025-05-22
Fixing inconsitent metabolic tasks

migp11:

[Metabolic_task_curation.xlsx](https://github.com/user-attachments/files/20395745/Metabolic_task_curation.xlsx)
Inconsistent tasks were manually curated by correcting metabolite identifiers. To curate infeasible tasks, we manually adjusted upper and lower bounds for the input and/or output metabolites or by updating the compartments of some metabolites. After the 
curation process, 19 metabolic tasks were reconciled to the Human-GEM model, whereas 8 tasks remained either inconsistent (1) or infeasible (7). A summary of the curation processes can be found in the attached table. Curated tasks were tested by computing a feasible flux distribution using pFBA (scripts used can be shared if needed).


mihai-sysbio:
It appears the diff is getting complicated to review here. To avoid messing up the branches, I'd recommend cherry-picking the relevant commits onto a branch that stems from `develop`.

Issue #1013, opened 2026-03-04
Reformulation of SAM synthesis GPR MAR03875 & possibly MAR07139


dagl1:
Both reactions (MAR03875 & MAR07139) have GPRs defined as MAT1 OR MAT2A OR MAT2B, however MAT1 functions as homodimer or homotetramer, whereas MAT2A and MAT2B form a complex (MAT2A as dimer + MAT2B), with MAT2A forming the catalytic subunit and MAT2B functioning as regulatory subunit, decreasing Km for methionine.
https://pmc.ncbi.nlm.nih.gov/articles/PMC3198898/
https://journals.physiology.org/doi/full/10.1152/physrev.00047.2011  

Evidence for selenomethionine to SeAdoMet conversion is a bit scarce: https://pmc.ncbi.nlm.nih.gov/articles/PMC1131364/ page 771, left bottom corner (indicating that AdoMet and SeAdoMet compete for their respective enzyme(s)). Thus it is unclear from this search whether SeAdoMet even is converted by both MAT1 and MAT2 or only a single complex.

Expected change: 
GPR of MAR03875 to go from MAT1 OR MAT2A OR MAT2B -> MAT1 OR (MAT2A AND MAT2B) 

(optionally, if deciding to keep both enzyme complexes for MAR07139): same GPR change as MAT2B is not a catalytic subunit.

edkerk:
You are right that the current GPR is wrong, and thanks for the references. We have refined the fix slightly on the basis of knockout data.

The current rule lets MAT2B catalyse the reaction on its own, which it cannot, since MAT2B is a non-catalytic regulatory subunit. Your `MAT1A OR (MAT2A AND MAT2B)` is closer, but it would make MAT2B required for the MAT2 route, and experimentally it is not: knocking out MAT2B leaves cellular SAM unchanged, while only MAT2A depletion lowers SAM ([Wan et al. 2024, Cell Death & Disease](https://pubmed.ncbi.nlm.nih.gov/39353892/)). MAT2A forms a catalytically active homotetramer without MAT2B ([Murray et al. 2014](https://pmc.ncbi.nlm.nih.gov/articles/PMC3793261/); [Xia et al. 2021, Biochemistry](https://pubs.acs.org/doi/10.1021/acs.biochem.1c00672)). MAT2A is therefore catalytic and self-sufficient, MAT1A is the separate catalytic isozyme, and MAT2B only modulates and stabilises MAT2A.

[...]

PR #1030, opened 2026-07-11
fix: reformulate SAM synthesis GPR and remove non-catalytic MAT2B

edkerk:
As reported in #1013, the GPR of MAR03875 (SAM synthesis) and MAR07139 (SeAdoMet synthesis) was `MAT2B or MAT1A or MAT2A`, which incorrectly lets the non-catalytic regulatory subunit MAT2B catalyse the reaction on its own.

- Set the GPR of MAR03875 and MAR07139 to `ENSG00000151224 or ENSG00000168906` (MAT1A or MAT2A).
- Remove MAT2B (`ENSG00000038274`) from the model, as it is non-catalytic and was associated only with these two reactions.

MAT2B knockout leaves cellular SAM levels unchanged while only MAT2A depletion reduces SAM (PMID:39353892), and MAT2A forms a catalytically active homotetramer without MAT2B, so MAT2B should not gate flux. This differs from the reporter's `MAT1A OR (MAT2A AND MAT2B)`, which would make MAT2B falsely essential for the MAT2 route.


edkerk:
Regarding the gene essentiality results: the model changes in this PR result in MAT2A predicted essential in DLD1, HELA, RPE1 and GBM.
- In HCT116 it remains non-essential (FN compared to Hart 2015 data) because MAT1A is expressed at 1.68 TPM, just above the 1 TPM ftINIT cutoff, so it is kept in the HCT116 model and covers SAM synthesis.
- In GBM, Hart 2015 does not call MAT2A an essential gene, which is likely a FN in the screen: 93.6% of genes strongly essential in the other four lines are also essential in GBM, but MAT2A is among the misses, alongside core genes such as MTOR and BRD4.

Issue #1020, opened 2026-05-01
prepHumanModelForftINIT fails with Human-GEM v2.0.0


mfcesur:
Hi,

The _prepHumanModelForftINIT_ function fails with Human-GEM v2.0.0 at: spontRxnNames = rxns_tsv.rxns(spont == 1);

This occurs because the _importTsvFile_ function reads the 'spontaneous' column in **reactions.tsv** as a cell array in v2.0.0, whereas older versions return numeric values.

As a result, the comparison (spont == 1) fails with: Operator '==' is not supported for operands of type 'cell'. This is due to changes in the reactions.tsv format in Human-GEM v2.0.0.

Replacing _importTsvFile_ with _readtable_ appears to resolve this issue: rxns_tsv = readtable(rxnsFilePath,'FileType','text','Delimiter','\t');

Best regards,  
Müberra

edkerk:
Thanks, confirmed, and your diagnosis is correct. From v2.0.0 the reactions.tsv is written without quoted fields, so `importTsvFile` returns the `spontaneous` column as text rather than numeric, and `spont == 1` then errors. In v1.x the file was quoted, so the same column was read as numeric.

`prepHumanModelForftINIT` will read the `spontaneous` column as numeric regardless of how the tsv is quoted, so it works on v2.0.0 and later. Your `readtable` workaround is fine in the meantime.

mfcesur:
Thank you for confirming the issue and fixing it.

PR #1031, opened 2026-07-11
fix: read spontaneous column as numeric in prepHumanModelForftINIT

edkerk:
As reported in #1020, `prepHumanModelForftINIT` fails on Human-GEM v2.0.0 at `spont == 1` with "Operator '==' is not supported for operands of type cell".

From v2.0.0 the reactions.tsv is written without quoted fields, so `importTsvFile` returns the `spontaneous` column as text rather than numeric (in v1.x the file was quoted, so it was read as numeric).

- Coerce the `spontaneous` column to numeric in `prepHumanModelForftINIT` when it is returned as text, so the function works regardless of how reactions.tsv is quoted.


edkerk:
No model changes.

Issue #1003, opened 2025-07-25
Pentose phosphate pathway subsystem appears twice


wshao1:
Hi,

It appears that the Pentose phosphate pathway subsystem appears twice in the model. Once as "Pentose Phosphate Pathway", containing only MAR20070. And once as "Pentose phosphate pathway", containing another set of reactions. I assume the two instances should be combined into one group?

edkerk:
Confirmed, thanks. As already noted above, the duplicate comes only from [MAR20070](https://metabolicatlas.org/explore/Human-GEM/gem-browser/reaction/MAR20070), whose subsystem was set to "Pentose Phosphate Pathway" (title case) when it was added, while the other 25 reactions of the pathway use "Pentose phosphate pathway". MAR20070's subsystem is corrected to "Pentose phosphate pathway", so the two groups merge into one.

PR #551, opened 2023-05-05
fix: sedoheptulose kinase/phosphatase reactions

Devlin-Moyer:
#### Main improvements in this PR:
Make changes proposed in #547:
- Remove MAR04568, since it doesn't appear to actually occur in humans
- Make MAR04567 irreversible in the direction of ATP consumption (and switch products and reactants so it can't only have negative fluxes)
- Add new reaction MAR20070 to represent the corresponding phosphatase reaction: MAM02883c + MAM02040c -> MAM02884c + MAM02751c

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Selected `develop` as a target branch
- [X] Any removed reactions and metabolites have been moved to the corresponding deprecated identifier lists

PR #1032, opened 2026-07-11
fix: merge duplicate pentose phosphate pathway subsystem

edkerk:
As reported in #1003, the pentose phosphate pathway appears as two subsystems: MAR20070 has its subsystem set to "Pentose Phosphate Pathway" (title case), while the other 25 reactions of the pathway use "Pentose phosphate pathway". The wrong casing was introduced when MAR20070 was added in #551.

- Correct the subsystem of MAR20070 to "Pentose phosphate pathway", merging the two groups into one.


edkerk:
No structural change

PR #1048, opened 2026-07-11
fix: update ferrochelatase (FECH) EC number to 4.98.1.1

### LINKED ISSUE #366: Changed EC number in other databases
### Description of the issue:
Protoporphyrin ferrochelatase enzyme (FECH) was classified as EC 4.99.1.1 but now it's classified as 4.98.1.1. This change is already updated in several databases including the Enzyme database, KEGG, MetaCyc. 
I came across this one accidentally, but I am wondering if there is any automatic way to do these checkings and update the model once in a while? 

We can update this one manually, for now. If you @mihai-sysbio will update it on the Metabolic atlas, I think we should keep the old EC number as well. 

- [ ] Done this analysis in the `main` branch of the repository
- [ ] Checked that a similar issue does not exist already

edkerk:
Resolves #366. Ferrochelatase (FECH) was reclassified in the IUBMB Enzyme nomenclature: EC 4.99.1.1 is now a transferred entry pointing to EC 4.98.1.1 (protoporphyrin ferrochelatase), part of the new sub-class 4.98 for enzymes that form carbon-metal bonds.

- Update `MAR01044` (Ferrochelatase, Mitochondrial) `eccodes` from `4.99.1.1` to `4.98.1.1`.

The old code is a deleted entry and is replaced rather than kept, to avoid a dead lookup for EC-based tools such as GECKO. Verified against the ExPASy ENZYME database.

### DIFF
--- a/data/testResults/README.md
+++ b/data/testResults/README.md
@@ -4,7 +4,7 @@ The file here contains results from the [MACAW](https://github.com/Devlin-Moyer/
 
 The test results shown here were obtained by the GitHub Actions run in:
 
-- **PR #1035** (QC)
+- **PR #1048** (QC)
 - **PR #973** (gene essentiality)
 
 The results will be updated by any subsequent pull request. Summary results are shown as a comment in the corresponding pull request.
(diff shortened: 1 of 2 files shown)


edkerk:
No structural model change

Issue #416, opened 2022-09-22
Many reactions have GPRs but not EC number. Is it possible to fill them in somehow?


johan-gson:
### Description of the issue:
I'm currently working with GECKO3, and one of the ideas there is to work with the EC field in the model directly instead of identifying the EC. For Human-GEM, many reactions with GPRs have an empty EC field. Perhaps we could use getEnzymeCodes in GECKO2 to fill them in? Or has this been tried already?

**I hereby confirm that I have:**
- [ ] Tested my code on my own computer for running the model
- [ ] Done this analysis in the `main` branch of the repository
- [X] Checked that a similar issue does not exist already

haowang-bioinfo:
sounds good. @IVANDOMENZAIN what do you think?

IVANDOMENZAIN:
I agree, this can easily be done

johan-gson:
Sounds good!

johan-gson:
I'm thinking we only replace the empty once to stay on the safe side, right?

johan-gson:
Another interesting thing - it seems that some reactions also have EC numbers but no GPRS - I'm thinking that filling in the GPRs for those may be possible? If we don't have the GPRs, ftINIT cannot use them, and GECKO will also have problems, at least if we want to use dlkcat.

mihai-sysbio:
> filling in the GPRs for those may be possible?

To me this sounds like a lot more work, and likely much less automatic unless one would use a tool such as GPRuler (which may take hundreds of hours on the cluster to run). From what I noticed, there isn't a general issue for GPR curation in this repo. My suggestion would then be to restrict the current issue to fill in missing ECs with GECKO, and open a new one for GPR curation.

johan-gson:
@IVANDOMENZAIN Have you had time to look at this? I'm just thinking that for GECKO3 to work properly with Human-GEM, this needs to be fixed. It sounds like a quick thing to do, or did you run into trouble?

edkerk:
Implemented a first pass at this. Two things became clear on closer look.

First, the gap is smaller than the raw count suggests. Of the reactions with a GPR but no EC, about 2000 are transport reactions, which are classified by TC number rather than EC and so correctly have none. A further set already carried EC numbers stored as a list rather than a single value. The real gap is 767 metabolic reactions.

[...]

PR #1051, opened 2026-07-11
fix: fill missing and correct obsolete EC numbers

edkerk:
EC number curation (#416, #366), entirely within `eccodes` fields, so model structure and balance are unchanged.

**Fill missing (#416).**
- 236 reactions with a GPR but no EC, from KEGG reaction mappings corroborated by gene-level EC (UniProt/KEGG via `genes.tsv`).
- A further 126 from the genes' EC: 94 by consensus across the already-annotated model (each UniProt-confirmed) and 32 taken directly from the genes' UniProt annotation. Assigned as a full code where the genes agree, or a sub-subclass wildcard (`a.b.c.-`) where they agree only to that level. Transport reactions and SLC/ABC genes are excluded as evidence.

**Correct existing (#366).** Checked against the IUBMB/ExPASy ENZYME database:
- 507 reactions where an EC had been transferred to a new number, mainly the 2018 reclassifications (ABC-transporter ATPases `3.6.3.x -> 7.6.2.x`, cytochrome P450s `1.14.13.x -> 1.14.14.x`).
- Typo and deleted-entry fixes: `1.4.11.1 -> 1.14.11.1`, `2.1.3.85 -> 2.3.1.85`, removed the spurious `5.2.2.2` from phosphoglucomutase, `3.1.2.15 -> 3.4.19.12`.
- 9 ambiguous cases left unchanged for manual review.

FECH (`4.99.1.1 -> 4.98.1.1`) is handled in #1048.


edkerk:
No structural model changes

Issue #181, opened 2020-06-17
missing reaction names


pecholleyc:
### Description of the issue:
A large amount of reactions in the model do not have a descriptive name.

#### Expected feature/value/output:
More reactions with descriptive names in the model.

#### Current feature/value/output:
8200+/13400+ reactions without name.

#### Reproducing these results:
search for  `    - name: ""\n      - metabolites` in the .yml

Most of the current reaction names in the model are identical to the BiGG or Recon3D annotation. But using the BiGG / Recon3D, KEGG and Reactome external identifiers I estimate that 3500+ additional reaction names could be imported in the model (based on v1.3). 

I think names then could also be curated or auto-generated by considering the equation and/or EC code of enzymes associated to the reactions.

**I hereby confirm that I have:**
- [X] Tested my code on my own computer for running the model
- [X] Done this analysis in the `master` branch of the repository
- [X] Checked that a similar issue does not exist already

Hao-Chalmers:
@pecholleyc nice to have this issue. 

This should be a long-term thing and it will take some time to fully resolve the reaction names. Might be good to begin with 1-2 external rxn ids for importing the names.

mihai-sysbio:
While I fully support the idea behind this issue, I don't have a straightforward suggestion here. It feels like there is no "ground truth" database to be used for the reaction names in a way that would resolve a majority of the empty names.

My opinion is that, if possible, this should be scripted in a way that it can be run repeatedly.

Hao-Chalmers:
Can't agree more

mihai-sysbio:
I'm at the point where I think _any_ names would be better than the 8000+ reactions with no names.

One way to do this would be to fetch the names in KEGG ([example](https://www.genome.jp/dbget-bin/www_bget?rn:R00711)).
Alternatively, the names can be fetched based on the E.C. code ([example](https://www.brenda-enzymes.org/enzyme.php?ecno=1.2.1.4)). That sounds more tricky since there are over 7600 empty `eccode`, and other entries with multiple E.C. cores.

Any thoughts?

mihai-sysbio:
I hope it's okay to ping @haowang-bioinfo and @JonathanRob to discuss the idea mentioned above: reactions in KEGG have names. Would it make sense to programmatically use KEGG as a source for reaction names?

haowang-bioinfo:
@mihai-sysbio do you have other suggested sources besides KEGG?

mihai-sysbio:
[...]

PR #1052, opened 2026-07-11
feat: add descriptive names to all reactions (100% coverage)

edkerk:
Addresses #181 by filling in a descriptive name for every reaction that had none. Reaction name coverage goes from 56% to 100% (all 12877 reactions named). The change is confined to the `name` field, so model structure and balance are unchanged, and the generation is scripted so it can be re-run as the model grows.

Names come from six sources, in priority order:
- **Transport reactions (2457)**: the transported metabolite and its compartments, e.g. `transport of GSH (cytosol to mitochondria)`.
- **EC number (1526)**: the ExPASy/IUBMB enzyme name for the reaction's EC plus the primary substrate, e.g. `carboxylesterase (substrate)` (transferred ECs resolved to the current entry first).
- **Single gene (934)**: the gene's enzyme name, trimmed of nomenclature suffixes like "family N member M", plus the primary substrate, e.g. `cytochrome P450 (ethanol)`.
- **Gene consensus (184)**: multi-gene reactions where the genes agree on one enzyme name (unanimously or by clear majority), plus the substrate.
- **Reaction chemistry (553)**: for reactions with no EC, gene agreement or KEGG name, the transformation is reconstructed from the stoichiometry in the style `<substrate> <transformation>`, e.g. `fructose phosphorylation` (ATP to ADP), `retinoate activation` (ATP to AMP + PPi + CoA), `<substrate> to <product> conversion` otherwise. These are derived from the reaction's own chemistry rather than an external database, so a few edge cases may be approximate.
- **KEGG systematic name (6)**: the few reactions whose KEGG entry provides a real name (equation-only entries skipped).

Only previously unnamed reactions were touched.

Follow-up commits also normalise problematic characters across all reaction names (collapse `--`/`---` to `-`, remove stray spaces left by ExPASy line-wrapping while keeping genuine elisions like "formate- and CO-forming", replace semicolons with commas, collapse double spaces, trim trailing punctuation and leading/trailing whitespace) and shorten the longest names so they roughly convey the reaction rather than fully specify it. Any name over ~80 characters is regenerated to a short form (gene or EC enzyme name, else a transport/exchange or chemistry-derived form), with long metabolite names taken to their KEGG glycan id where present or truncated. The longest name drops from 324 to 90 characters (median 45). This deliberately replaces some official systematic names with rougher, shorter ones.


edkerk:
No structural model changes

